PR #27, opened 2022-02-15
Reecan/test add mutaflag

so-dopamine:


shaoqx:
Thank you for writing the test for Add_MutaFlag. As a practice, we will not merge these changes. One comment: test_flag_is_string() is not testing the output of Add_MutaFlag.

PR #29, opened 2022-02-15
Reecan/readme update

so-dopamine:
minor changes to readme file.


shaoqx (approved):
Thanks for editing the README! This is great! One minor thing is we always want to PR to the develop branch first. But in this case, the README does not involve any code in it and the revision is in urgent need - we want to show it to users before the next complete PR from develop. So we will skip the develop branch only this time.

PR #33, opened 2022-02-22
Fix Docstring for Add_MutaFlag

mtremblay0817:
Update Add_MutaFlag docstring to fit Google style guide

### DIFF
--- a/Class_PDB.py
+++ b/Class_PDB.py
@@ -3,6 +3,7 @@ import os
 import re
 from subprocess import run, CalledProcessError
 from random import choice
+from xmlrpc.client import boolean
 from AmberMaps import *
 from wrapper import *
 from Class_Structure import *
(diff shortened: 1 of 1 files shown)


Class_PDB.py
```diff
@@ -860,31 +861,29 @@ def PDB2PDBwLeap(self):
...
+        Checks assigned mutations, generate mutation randomly based on rules,
+        and calculate mutation based on rules.
+
+        Args:
+            Flag: str or list of strings indicating mutation to perform (see below)
+            if_U: if selenocysteine (U) is allowed for random mutation
+            if_self: if mutation to the same amino acid is allowed for random mutation
+        
+        Returns:
+            A str representing the current mutations to the PDB.
```
shaoqx, line 876: return should be a list

Class_PDB.py
```diff
@@ -860,31 +861,29 @@ def PDB2PDBwLeap(self):
...
+        
+        Flag Grammar:
```
shaoqx, line 878: the part should be inside of the Args section

shaoqx (changes requested):

shaoqx (approved):

PR #34, opened 2022-02-22
Create rJedit_Add_MutaFlag_doc.py

so-dopamine:
EnzyHTP agenda 2022/02/22

### DIFF
--- a/fix/rJedit_Add_MutaFlag_doc.py
+++ b/fix/rJedit_Add_MutaFlag_doc.py
@@ -1,0 +1,79 @@
+def Add_MutaFlag(self,Flag = 'r', if_U = 0, if_self=0):
+        """Returns a label of mutations.
+
+        Reads input of single string or list of strings. Generates a random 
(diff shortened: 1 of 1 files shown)


fix/rJedit_Add_MutaFlag_doc.py
```diff
@@ -0,0 +1,79 @@
...
+                A: chain index.
+                11: residue index.
+                Y: target residue mutation.
+
+            if_U: include mutations to U (selenocysteine) in random generation 
+            if_self: for random generation, compares mutation to original 
+              residue.
+
+        Returns:
+            A list of strings that represent the original residue letter,
```
shaoqx, line 24: talk about real return. And also the self.MutaFlags contains a list of tuples that contains strings inside of them. It's also useful to give examples.

shaoqx: The idea is to make sure others know how to use these outputs without reading the code.

shaoqx (changes requested):

PR #37, opened 2022-02-22
Xinchun/fix add muta flag anot

Xinchunran:
update the docstring on fix add mutaflag


Class_PDB.py
```diff
@@ -861,25 +861,32 @@ def PDB2PDBwLeap(self):
...
-        Grammer:
-        X : Original residue name. Leave X if unknow. 
+        '''generate the mutation letter  and add mutations to PDB.
+        
+        Append Flags on the self.MutaFlags list with mutation residues letter from
+        randomly assigned or from the manually assigned 
+        
+        Args:  
+            Flag   :A string or A list of string indicate the mutation residues
+            on PDB files.
```
shaoqx, line 871: flag grammer here

Class_PDB.py
```diff
@@ -861,25 +861,32 @@ def PDB2PDBwLeap(self):
...
+            if_U   :if U can be used for random generations (Selenocysteine).
+            Default 0
+            if_self:if the same amino acids can be used  in random
+            generation(wildtype)
+        
+
+        Return:
```
shaoqx, line 878: revise return

shaoqx (changes requested):

shaoqx:
Thank you for editing the doc string for Add_MutaFlag. As a practice we wont apply those changes. Also make sure that there's only one feature on each feature branch.

PR #40, opened 2022-03-15
new feature: MutaFlag_san_check

so-dopamine:


shaoqx:
Will get back to this after the refactoring

PR #42, opened 2022-03-15
new feature: refactoring decode MutaFlag

alvinnebula:
Update decode function


shaoqx:

PR #43, opened 2022-03-29
Reecan/feature muta flag san check

so-dopamine:
defined resi_1, resi_2, and other variables. added check to see if input residue matches original structure residue. added switch for resi_2.


shaoqx:
Note that we should always PR to develop and only develop can PR to master.

PR #47, opened 2022-04-12
Xinchun/random_generate_mut

Xinchunran:
add the function for the random generation and distance functions


shaoqx:

PR #50, opened 2022-04-18
BugFix: fix PDB2FF() error writing the same file

shaoqx:
fix PDB2FF() error writing the same file when multiple jobs work in local_lig=0 mode.
A file IO error will show up when multiple tasks are generating the same ligand-related files at the same time.


shaoqx:
@KleinesMesser Please describe details about this problem and test it with jobs.

shaoqx:
This problem still exists if running multiple mutations in parallel on a cluster. If one job starts next to the 1st one without the 1st finishing PDB2FF(), it will still re-create those ligand-related files.

PR #72, opened 2022-07-24
Xinchun/conf equi json loader

Xinchunran:
update the conf_equi_json file & purest on the function


chrisjurich:
I will not be approving this PR at the moment. There are a few issues that need to be fixed before bringing it in:

- [ ] I think you need to re-base this branch to a newer version of develop_refactor. This one is missing the implementation of AmberConfig, which has already been created.
- [ ] you need to use the core._LOGGER singleton instance that already exists instead of making a new one
- [ ] When you make tests that use files, use the __file__ attribute to get the path correct in a more robust manner

shaoqx:
Also please format the committing message as:
[Type]: [a concise message]
This is from the need of **searching** for a specific code change in the future. 
e.g.: Update: xxxx, or Bugfix: xxxx, etc. The message itself can be vague but at least we can have a type.

shaoqx:
Thank you for the change! As a practice, we will not merge these changes.

PR #73, opened 2022-07-24
submit refactor assignment of run_command() and its test

alvinnebula:
**_Added behavior:_**

If any argument in the list of `args` doesn't entirely consist of special characters, wrap it in quotes.

_Create a draft pull request for review._


shaoqx:

PR #75, opened 2022-07-27
Shaoqz/refactor structure attr mod

shaoqx:
In this PR:
- several hotfix for a working development environment
- fixed some bug from last PR #74 
- refactor about Structure attr. (mostly how to deal with the duplicated chain name)
   + is there a better way to set the coordinate level comparison of chains?


chrisjurich (approved):
Good work overall. I really like the use of @property throughout. I will work on fixing the installation requirements.

PR #76, opened 2022-07-28
Bugfix: raise a bug of PDB w/o chain id & w/ANISOU

shaoqx:
Current method of dealing with missing chain id is to detect residues that are not neighbors in terms of lines in the PDB file.

This would work in most time since the only non-atom line in PDB most time will be TER and END which fits the case. But also in a lot of cases, PDBs from the PDB database contain [ANISOU records](http://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ANISOU) that separate every ATOM line, which I think will be the only exception.


chrisjurich (approved):
Appropriate test case

chrisjurich:
Thanks for creating this bug test case. The ANISOU record is something important to consider.

PR #78, opened 2022-07-29
Paul/conf prod json loader

KleinesMesser:
Added the function load_conf_prod() to amber_config.py.

There is an issue loading module Frame in the pytest file, so I did not update this file.


shaoqx:

PR #81, opened 2022-08-09
Xinchun/conf json loader updated

Xinchunran:
update the conf_equi_json_load following reviews comment


shaoqx:

Comments: 

- This branch is merging into the wrong branch.
- In the case you want to update an existing PR addressing the comments, you should just commit to the same branch in the old PR instead of making a new one with a new branch.

PR #85, opened 2022-10-06
Shaoqz/refactor protonation&clean

shaoqx:
This PR contains the refactoring of the rm_water() and get_protonation() function is original EnzyHTP. These two functions are now 2 individual submodules under enzy_htp/preparation: protonation.py and clean.py. Generic API applying changes to Structure() is designed as science APIs (under preparation) and general APIs (under structure_operation). APIs in general.py is waiting for further division into different files in the future.
The structure module is also cleaned and refactored along the way in those clean/protonation-related parts.
Fixed a major bug in structure_io class about chain id fix for HETATM chains.


enzy_htp/core/exception.py
```diff
@@ -39,3 +32,11 @@ class InvalidMutationRestriction(Exception):
     """Exception corresponding to an invalid mutation restriction in enzy_htp.mutation.mutation_restrictions.py."""
 
     pass
+
+class ResidueDontHaveAtom(Exception):
```
chrisjurich, line 36: This doesn't read correctly in terms of english grammar. `ResidueMissingAtom(Exception)` would make more sense.

enzy_htp/core/file_system.py
```diff
@@ -19,12 +19,23 @@ def safe_rm(fname: str) -> None:
     if os.path.exists(fname):
         os.remove(fname)
 
-
-def safe_rmdir(dirname: str) -> None:
+def is_empty_dir(dir_path: str) -> bool:
+    '''
```
chrisjurich, line 23: Double quotes instead of single quotes.

enzy_htp/core/math_helper.py
```diff
@@ -0,0 +1,85 @@
...
+from .logger import _LOGGER
+import numpy as np
+import math
+
+def check_valid_ph(ph: float) -> None:
+    """Helper function that checks the pH is on the range: [0.00, 14.00]. Give a warning if not."""
+    if ph < 0 or ph > 14:
+        _LOGGER.warning(f"assigned pH: {ph:.2f} out of range: [0.00,14.00]")
+
+def set_distance(p1,p2,d):
```
chrisjurich, line 15: Missing type hinting, doc string does not resemble google style guide for python. Also, please use double quotes, not single.

enzy_htp/core/math_helper.py
```diff
@@ -0,0 +1,85 @@
+"""Module holds functions for math calculations.
```
chrisjurich, line 1: I won't annotate every problem with every functions in this file, but please add the type hinting as well as standardized docstrings.

enzy_htp/structure/atom.py
```diff
@@ -6,11 +6,17 @@
...
 import sys
-from typing import Tuple
+from typing import List, Tuple
+from plum import dispatch
 
 import numpy as np
 import pandas as pd
+
+import enzy_htp.chemical as chem
+from enzy_htp.core.exception import ResidueDontHaveAtom
```
chrisjurich, line 18: Try to rename this Exception to make more grammatical sense.

enzy_htp/structure/atom.py
```diff
@@ -100,23 +109,136 @@ def b_factor(self, val):
...
+        return self._connect is not None
     #endregion
 
     #region == Special ==
     def __str__(self):
         return f"Atom({self._name}, {self._idx}, {self._coord}, {self._parent}, {self._b_factor}, {self._element}, {self._charge} )"
     #endregion
 
+    @dispatch
+    def _(self):
```
chrisjurich, line 239: What is this?

enzy_htp/structure/chain.py
```diff
@@ -105,6 +120,11 @@ def num_residues(self) -> int:
         """Returns number of Residue() or Residue()-dervied objects belonging to the Chain."""
         return len(self)
 
+    @property
+    def residue_idxes(self) -> List[int]:
```
chrisjurich, line 124: probably rename this to `residue_idxs()`

enzy_htp/structure/ligand.py
```diff
@@ -62,34 +62,22 @@ def is_ligand(self) -> bool:
...
+        """
+        name_list = self.atom_name_list
+        new_name_list = chem.get_valid_generic_atom_name(name_list)
+        for name, atom in zip(new_name_list, self.atoms):
+            if atom.name != name:
+                _LOGGER.info(f"found atom with invalid name {atom}. changing it to {name}")
+                atom.name = name
 
     # === Special ===
     def __str__(self) -> str:
```
chrisjurich, line 78: Very elegant! I like this.

enzy_htp/structure/residue.py
```diff
@@ -55,7 +55,15 @@ def atoms(self) -> List[Atom]:
     @atoms.setter
     def atoms(self, val):
         self.set_children(val)
-
+    
+    @property
+    def _atoms(self) -> List[Atom]:
+        """alias for _children. Prevent change children but _atoms dont change"""
+        return self._children
+    @_atoms.setter
```
chrisjurich, line 63: Add a space here between this and the above method.

enzy_htp/structure/structure.py
```diff
@@ -212,6 +291,59 @@ def has_duplicate_chain_name(self) -> bool:
             existing_c_id.append(ch.name)
         return False
     
+    def is_idx_subset(self, target_stru: Structure) -> bool:
+        """
```
chrisjurich, line 295: Please update the docstring so that it is up to standard with the google style guide for python with respect to docstrings

chrisjurich (changes requested):
Overall I think this merge is pretty well polished. Would just like to see the renaming of the Missing atom exception and also to have the correct linter run on the source code. Very good test coverage though.

enzy_htp/core/exception.py
shaoqx: But I think missing means it should have right? In this case, we do not know if the query atom name is supposed to be carried by the residue. What could be a better name in this case?

enzy_htp/structure/atom.py
shaoqx: This is required by plum-dispatch when use it in a class

enzy_htp/structure/ligand.py
shaoqx: Thanks! Learned from reading your code!

enzy_htp/structure/residue.py
shaoqx: Not having space here was because those getter/setter methods are meant to be in pair. I see treat them here together as a "method".

enzy_htp/core/exception.py
chrisjurich: Ah I understand. In that case `ResidueDoesntHaveAtom(Excpetion)` would be more fitting.

chrisjurich (approved):
Looks good!

shaoqx:
Cool! Thank you for the review!

PR #87, opened 2022-10-13
Chrisjurich/fix chemdb

chrisjurich:
This PR includes a number of changes aimed at removing the existing setup for chemical database accession as well as moving all interface and config classes to the respective `_interface` and `_config` directories. Additionally, have cleaned up unit tests so that all but three currently pass.


shaoqx:
Thank you for the changes! Please branch from develop_refactor.

PR #89, opened 2022-10-15
Chrisjurich/remove chem db

chrisjurich:
Removing the last traces of the `load_from_db()` calls from the `enzy_htp.chemical.db` submodule. Tests/installation should be possible now.


enzy_htp/chemical/residue.py
```diff
@@ -33,46 +230,5916 @@
 
 RESIDUE_CONNECTIVITY_MAP: Dict[str,
                                Dict[str,
-                                    List[str]]] = load_from_db("RESIDUE_CONNECTIVITY_MAP")
+                                    List[str]]] = {
+        "ALA": {
```
shaoqx, line 234: Is it possible to not make it so long? the original format looks fine to me

shaoqx: Oh just found out it was like this in the json. Maybe you can refer to AmberMaps

enzy_htp/chemical/residue.py
```diff
@@ -6,20 +6,217 @@
...
+        "MET": "M",
+        "PHE": "F",
+        "PRO": "P",
+        "SEC": "U",
+        "SER": "S",
+        "THR": "T",
+        "TRP": "W",
+        "TYR": "Y",
+        "VAL": "V"
+} #TODO(shaoqz): @imp2 add related to canonical in the name
```
shaoqx, line 61: I added several alternative names for some residues. Please add it back

test/chemical/test_residue.py
```diff
@@ -15,12 +15,12 @@ def test_variable_checks():
     assert len(res.AA_LIST) == 21
     assert all_caps(res.AA_LIST)
 
-    assert len(res.THREE_LETTER_AA_MAPPER) == 29
```
shaoqx, line 18: It should be 29 after those added names are back

shaoqx (changes requested):
Nice! Just one minor change is requested.

enzy_htp/chemical/residue.py
chrisjurich: Updated this

enzy_htp/chemical/residue.py
chrisjurich: Added extra codes back in.

test/chemical/test_residue.py
chrisjurich: Updated.

shaoqx (approved):
Good job!

PR #103, opened 2023-02-21
AmberInterface Updates and Electric Field implementation-

chrisjurich:
In this PR I include updates to the AmberInterface class so that it can now parse prmtop files and add charges to existing Structure objects. I also present an implementation of the electric_field() function. This function DOES NOT support the ability to use masks to select the enzyme atoms but will be implemented in the future.


enzy_htp/_config/pymol_config.py
```diff
@@ -0,0 +1,51 @@
+"""Defines PyMOLConfig() which holds configuration settings for enzy_htp to interface with the
```
shaoqx, line 1: I also made a pymol interface. This will cause conflict when I merge it. Maybe we can merge our code manually first? Or any better ways to do this

enzy_htp/analysis/electric_field.py
```diff
@@ -0,0 +1,82 @@
+"""Submodule contains code for calculating the electric field strength in an enzyme system along a specified
```
shaoqx, line 1: The current function seems only designed to be working on a single structure. In real cases, EF are also calculated from trajectories. I will work on design classes to contain trajectory/ensemble (maybe a class hold structures: List[Structure]?) and maybe we can add a dispatch to this function in the future.

enzy_htp/structure/structure_io/pdb_io.py
```diff
@@ -643,7 +643,9 @@ def _write_pdb_atom(atom: Atom) -> str:
         insert_code = f"{'':1}"
         occupancy = f"{1.00:>6.2f}"
 
-        temp_factor = f"{atom.b_factor:>6.2f}"
+        temp_factor: str = ' ' * 6
```
shaoqx, line 646: Here is a fix in my branch:
```
        if atom.b_factor is None:
            temp_factor = f"{0.0:>6.2f}"
        else:
            temp_factor = f"{atom.b_factor:>6.2f}"
```

shaoqx: Maybe we update it here so that there wont be conflict?

enzy_htp/__init__.py
```diff
@@ -36,11 +37,16 @@
...
 
 from ._interface import Interface
 
+
 interface = Interface(config)
 
-#TODO(CJ): add docstrings for global variables
+from .analysis import ( electric_field )
+
+"""Singleton interface for all softwares enzy_htp is capable of interfacing with. SHOULD NOT be used by
```
shaoqx, line 50: This should be under interface?

test/preparation/test_protonate.py
```diff
@@ -225,3 +225,18 @@ def test_pybel_protonate_pdb_ligand_4CO():
     assert equiv_files(out_ligand_path, answer_ligand_path, skip_frist=True)
     fs.safe_rm(out_ligand_path)
     assert not os.path.isdir(out_ligand_path)
+
+
+def test_protonate_stru():
```
shaoqx, line 230: This is a duplicated name

shaoqx (changes requested):
Great PR! I noted some minor problems I noticed and some comments about future plans.
Also I will work on bring job_manager to AmberInterface and GaussianInterface in the near future FYI.

enzy_htp/analysis/electric_field.py
shaoqx: No changes needed here

enzy_htp/_config/pymol_config.py
shaoqx: I will add diffs in my branch.

enzy_htp/_interface/pymol_interface.py
```diff
@@ -0,0 +1,118 @@
+"""Defines a PyMOLInterface() class that serves as a bridge for enzy_htp to leverage the functionality
```
shaoqx, line 1: I will add diffs in my branch.

test/_config/test_pymol_config.py
```diff
@@ -0,0 +1,18 @@
+"""Testing the enzy_htp._config.PyMOLConfig class.
```
shaoqx, line 1: I will add diffs in my branch.

shaoqx (approved):
Great! This PR is now good to merge

PR #105, opened 2023-02-23
Prepping for a develop_refactor beta release

chrisjurich:
Just needed to remove a few files and change a few things so that the beta release will work in conda.

PR #106, opened 2023-02-24
Shaoqz/fix pr 105

shaoqx:
Fix duplicated commits in PR #105 by reverting the whole merge and cherry-picking wanted commits.
Also, fixed some non-repeating but deprecated changes in those reverted commits mainly related to using the distance_to method.


chrisjurich (approved):
Good job

PR #107, opened 2023-02-28
Shaoqz/refactor mutation api

shaoqx:
- Mainly finished `mutation_pattern` submodule that designs a pattern to specify mutants.
- Fixed many bugs along the way.
- Created the backbone for the new mutation module, still need to finishes several functions.
- Support pymol2 package in `PyMolInterface`
- Add structure_selection submodule for selecting part of the structure using pymol-based pattern
- Add a template for docstring of Science APIs


enzy_htp/chemical/atoms.py
```diff
@@ -9,7 +9,7 @@
 X_H_BOND_LENGTH: Dict[str, float] = {
         "C": 1.07,
         "N": 1.0
-}
+} # yapf: disable
```
chrisjurich, line 12: Get rid of this comment.

enzy_htp/core/general.py
```diff
@@ -0,0 +1,133 @@
...
+    """convert the result of get_interval_from_list() to a str"""
+    result = get_interval_from_list(target_list)
+    result = ",".join([f"{x[0]}-{x[1]}" if x[0] != x[1] else f"{x[0]}" for x in result])
+    return result
+
+
+def get_random_list_elem(traget_list: list):
+    """Helper method that randomly chooses an element from a list. numpy.random.choice() doesn't 
+    like to take elements from list()'s of tuples so this is the work around."""
+    return traget_list[np.random.randint(len(traget_list))]
```
chrisjurich, line 72: Use np.random.choice() and add a `seed` argument

enzy_htp/core/general.py
```diff
@@ -0,0 +1,133 @@
+"""This module contains helper functions related to general python/python library features
```
chrisjurich, line 1: This whole file needs better doc strings

chrisjurich (approved):
Looks really good. Documentation could be improved but we will take care of that in a later PR.

enzy_htp/chemical/atoms.py
shaoqx: Yapf will make it into one line if so. I want it to stay what it is rn. I searched about it and this is the solution they give https://github.com/google/yapf/issues/359

enzy_htp/core/general.py
shaoqx: I think seeding is a global behavior. That's the reason I didn't add `seed` for all random related functions.
I will change the np function

shaoqx:
Cool! I revised those based on the comment!

PR #115, opened 2023-07-07
Merging `RosettaInterface()`, updated `PyMolInterface()` and the ability to check for python modules within the `EnvironmentManager()`

chrisjurich:
This PR has three main components which are listed below:

1. Initial implementation of `RosettaInterface()`

A simple implementation of Rosetta within the EnzyHTP package. Currently offers support for 

- Relaxation (unconstrained, specified constraints, and loop-only constraints)
- Using `rosetta_scripts` in a generic manner with specified protocols
- Parameterizing new ligands/residues

2. Update of `PyMolInterface()` to only use `pymol2` and have a session-based approach.

Updated `PyMolInterface()` so that it has a default session (attribute `self.session_`) which it uses by default 
for each function. If the user chooses, they can create a session from the function `PyMolInterface.new_session()`
such that each function uses the state available in the session object. The `PyMolInterface()` also checks if `pymol2` is in the current environment and errors/exists if it is not.

3. Ability to check for python modules within `EnvironmentManageer()`

Using the `importlib` python module, EnzyHTP now looks for whether python modules exist in their environment. Each `<Package>Interface()` can now see if any modules are missing.

As before, the documentation for much of this code is in progress. Will be updated in a major commit some day.


enzy_htp/_interface/amber_interface.py
```diff
@@ -541,76 +503,58 @@ def md_run(self, prmtop: str, inpcrd: str, work_dir: str, mode: str = "CPU") ->
 
         return prod_nc
 
-    def run_tleap(self,
```
shaoqx, line 544: Somehow it changes this function back. Please keep run_tleap() and discard mutate() and renumber_pdb()

enzy_htp/_interface/amber_interface.py
```diff
@@ -831,67 +771,11 @@ def add_charges(self, stru: struct.Structure, prmtop: str) -> None:
...
             aa.charge = c
 
         if not stru.has_charges():
-            _LOGGER.error(
-                "Not all atoms in the supplied structure were assigned charges. Exiting..."
-            )
+            _LOGGER.error("Not all atoms in the supplied structure were assigned charges. Exiting...")
             exit(1)
             pass
-
```
shaoqx, line 846: Please also bring this function back. And in the best case, don't create a git records like deleting and adding the same thing again.

(I wonder why git works like this here. Shouldn't those new commits that on develop_refactor but not on your branch also applied to the final version?)

enzy_htp/_interface/bcl_interface.py
```diff
@@ -0,0 +1,122 @@
...
+        self.env_manager_.run_command(self.config_.BCL_EXE, flags)
+
+    def generate_conformers(self, molfile: str, n_conformers: int = 100, outfile: str = None) -> str:
+        """Method that creates conformers using the 'molecule:ConformerGenerator' method from the BCL.
+        Note that the supplied molfile must exist but outfile does not. All conformers are saved to one
+        file. When the outfile is not supplied, the output file is saved to the same directory as the 
+        input file with the suffix '_conformers' added to the name.
+
+        Args:
+            molfile: str() with the name of the input file containing the small molecule of interest.
```
shaoqx, line 62: What file formats are supported here? Would be good to add in the docstring

enzy_htp/_interface/bcl_interface.py
```diff
@@ -0,0 +1,122 @@
...
+            f"{int(n_conformers)}",
+            "-conformers_single_file",
+            f"{outfile}",
+        ]
+
+        self._run_bcl_cmd("molecule:ConformerGenerator", flags)
+
+        return outfile
+
+    def formal_charge(self, molfile: str) -> int:
```
shaoqx, line 95: Maybe get_formal_charge is a better name?

enzy_htp/_interface/moe_interface.py
```diff
@@ -0,0 +1,109 @@
...
+
+        fs.check_file_exists(input_svl)
+
+        if Path(input_svl).suffix != '.svl':
+            _LOGGER.error(f"The supplied svl file '{input_svl}' does not have the extension .svl. Exiting...")
+            exit(1)
+
+        self.env_manager_.run_command(self.config_.MOE_BATCH, ["-script", str(input_svl)])
+
+    def protonate(self, molfile: str, outfile: str = None, pH: float = 7.0) -> str:
```
shaoqx, line 62: Cool interface! We should also write an engine mimicing protonate_ligand_with_pybel() in preparation.protonate.protonate_ligand_with_moe() taking enzy_htp.Structure as input and make changes in place.

shaoqx: This change is not necessary for this PR tho

enzy_htp/core/typing.py
```diff
@@ -0,0 +1,56 @@
...
+
+    elif strategy == 'exception':
+        raise TypeError(err_msg)
+
+    else:
+        _LOGGER.error(f"The supplied exit strategy '{strategy}' is not valid. Allowed are 'exit' and 'exception'. Exiting...")
+        exit( 1 )
+
+
+def check_var_type( var:Any, target_type:Any, fail_strategy:str='exit') -> None:
```
shaoqx, line 38: Is this designed to replace thos "if isinstance"?

enzy_htp/mutation/engine.py
```diff
@@ -226,6 +230,72 @@ def _mutate_pymol(pdb: str, outfile: str, mutations: List[Mutation]) -> None:
     pass
 
 
+def _mutate_rosetta(pdb:str, outfile:str, mutations: List[Mutation]) -> None:
```
shaoqx, line 233: This file is deprocatced. Maybe find a way to move this function to api.py and mimic the format of mutate_stru_with_tleap().

(May be do this after merging is easier?)

enzy_htp/structure/atom.py
```diff
@@ -233,15 +233,18 @@ def radius(self, method: str = "ionic") -> float:
...
-    def distance_to(self, point: Atom) -> float:
-        """Get the distance to the other atom or a point."""
-        return mh.get_distance(self.coord, point.coord)
+    #@dispatch
+    #def distance_to(self, point: Atom) -> float:
+    #    """Get the distance to the other atom or a point."""
+    #    return mh.get_distance(self.coord, point.coord)
 
-    @dispatch
+    #@dispatch
```
shaoqx, line 241: Did you make this from an error from plum that says "float" is not achieved. I dont really understand that error and it disappears after it tried it the 5th time. like wtf.

test/mutation/mutation_pattern/test_position_pattern.py
```diff
@@ -20,19 +20,17 @@ def test_decode_position_pattern():
     test_stru = sp.get_structure(test_pdb)
     test_pattern = "resi 254 around 4"
 
-    assert set(m_p.decode_position_pattern(test_stru, test_pattern)) == set([
```
shaoqx, line 23: yapf is actually making them looks worse lol. Can we prevent this?

test/mutation/test_api.py
```diff
@@ -16,61 +16,47 @@
...
-        })
+    mutants = mapi.assign_mutant(test_stru, test_mutation_pattern, chain_sync_list=[("A", "B")], chain_index_mapper={"A": 0, "B": 450})
     print(mutants)
 
+
 def test_sync_mutation_over_chains():
     """test function works as expected"""
-    test_mutants = [
-        [Mutation(orig='ARG',target='ALA',chain_id='A',res_idx=3),
-         Mutation(orig='ARG',target='TRP',chain_id='A',res_idx=4)],
```
shaoqx, line 54: the reformating here is lol. Can we prevent this?

test/mutation/mutation_pattern/test_target_aa_pattern.py
```diff
@@ -23,17 +23,16 @@ def test_decode_target_aa_pattern_not_expr():
     test_pattern = "all not self"
     test_orig_resi = "ARG"
     assert set(m_p.decode_target_aa_pattern(test_orig_resi, test_pattern)) == set([
-        "GLU", "ASN", "HIS", "SER", "VAL", "GLN", "TYR", "PRO", "ASP", "CYS", "GLY",
```
shaoqx, line 26: Maybe an option prevent it from making shorter lines longer?

test/mutation/test_mutation.py
```diff
@@ -19,18 +19,12 @@
 
 def test_generate_from_mutation_flag_wt(caplog):
     """test if the function behave as expected on WT"""
-    assert mut.generate_from_mutation_flag("WT") == mut.Mutation(orig=None,
```
shaoqx, line 22: Same as above

test/preparation/test_protonate.py
```diff
@@ -45,14 +42,11 @@ def equiv_files(fname1: str,
 
 
 ligand_answer_list_1Q4T = [
-    "N1A", "C2A", "N3A", "C4A", "C5A", "C6A", "N6A", "N7A", "C8A", "N9A", "C1D", "C2D",
```
shaoqx, line 48: Same as above

test/preparation/test_protonate.py
```diff
@@ -147,15 +143,14 @@ def test_protonate_peptide_with_pdb2pqr_metal():
 
     stru = sp.get_structure(test_pdb)
     prot.protonate_peptide_with_pdb2pqr(stru, 7.0, int_pdb, int_pqr)
-    assert list(
```
shaoqx, line 150: Same as above

test/structure/structure_io/test_pdb_io.py
```diff
@@ -262,16 +256,13 @@ def test_resolve_missing_chain_id_repeat_with_multi_in_HET():
     test_mdl = f'{DATA_DIR}1Q4T_atom_res_ch_build_test.pdb'
     test_mdl_pdb = PandasPdb()
     test_mdl_pdb.read_pdb(test_mdl)
-    target_df = pd.concat((test_mdl_pdb.df['ATOM'], test_mdl_pdb.df['HETATM']),
```
shaoqx, line 265: same as above

test/structure/structure_selection/test_general.py
```diff
@@ -22,6 +22,4 @@ def test_select_stru():
     test_stru = sp.get_structure(test_pdb)
     test_pattern = "resi 254 around 4"
     select_obj = select_stru(test_stru, test_pattern)
-    assert sorted(select_obj.involved_residue_idx()) == [
```
shaoqx, line 25: same as above

test/structure/test_chain.py
```diff
@@ -41,15 +42,15 @@ def test_residue_idx_interval():
...
 
+
 def test_is_same_coord():
     """Testing that the Chain.is_same_coord() method works correctly"""
     pdb_file_path = f"{DATA_DIR}two_chain.pdb"
     stru: Structure = sp.get_structure(pdb_file_path)
     ch1, ch2 = deepcopy(stru.chains[0]), deepcopy(stru.chains[0])
     assert ch1.is_same_coord(ch2)
     assert ch2.is_same_coord(ch1)
-    ch2.atoms[0].coord = (ch2.atoms[0].coord[0] + 0.05, ch2.atoms[0].coord[1],
```
shaoqx, line 51: same as above

test/conftest.py
```diff
@@ -10,11 +12,7 @@ def lines_from_file(fname):
         return result
 
     @classmethod
-    def equiv_files(cls,
```
shaoqx, line 13: same as above

enzy_htp/_interface/pymol_interface.py
```diff
@@ -110,70 +227,266 @@ def config(self) -> PyMolConfig:
...
+        session.cmd.delete( 'all' )
+        session.cmd.load( molfile )
+        session.cmd.remove( 'hydrogens' )
+        session.cmd.save( new_file )
+
+        session.cmd.delete('all')
+
+        return new_file
+
+    def rename_atoms(self, template: str, in_file: str) -> str:
```
shaoqx, line 271: What is this? Could it be used to address [#117](https://github.com/ChemBioHTP/EnzyHTP/issues/117)

enzy_htp/_interface/pymol_interface.py
```diff
@@ -110,70 +227,266 @@ def config(self) -> PyMolConfig:
...
-
-    #     self.cmd.delete('all')
-    #     self.cmd.load(fname)
-    #     lines: List[str] = self.cmd.get_fastastr(sele).splitlines()
-    #     self.cmd.delete('all')
-
-    #     lines = list(filter(lambda ll: ll[0] != '>', lines))
-    #     return ''.join(lines)
+
+    def de_protonate(self, molfile : str, new_file:str=None, session:pymol2.PyMOL=None) -> str:
```
shaoqx, line 231: If this is used for a protein. I think is will have the same issue as [#118](https://github.com/ChemBioHTP/EnzyHTP/issues/118)

enzy_htp/_interface/pymol_interface.py
```diff
@@ -110,70 +227,266 @@ def config(self) -> PyMolConfig:
...
+        session.cmd.iterate_state(state,sele,iter_stmt,space=_eh_local)
+
+        result = pd.DataFrame(
+            data=_eh_local['data'],
+            columns=variables
+        )
+
+        return result
+
+    def execute(self, args:List[Tuple], session:pymol2.PyMOL=None) -> List[Any]:
```
shaoqx, line 353: Could general_cmd() a better name. So that it is comparable with other functions in the interface that runs specific cmds.

enzy_htp/_interface/pymol_interface.py
```diff
@@ -8,77 +8,193 @@
...
+        Args:
+            file_1 : The name of the original file as a str().
+            file_2 : The name of the output file as a str(). Optional.
+            new_ext : The new extension to use. Optional.
+            session : A pymol2.PyMOL() session to use. By default uses an internal session instance that resets each function call.
+
+        Returns:
+            The name of the new file as a str().
+        """
+        self.check_pymol2_installed()
```
shaoqx, line 74: why is this check exist for all methods? Could having it in __init__ so it checks when the singletion object spawns?

enzy_htp/_interface/pymol_interface.py
```diff
@@ -8,77 +8,193 @@
...
+            sele: The str() specifying the atom selection in PyMOL synatx. Default is '(all)'.
+            session : A pymol2.PyMOL() session to use. By default uses an internal session instance that resets each function call.
+        """
+        pass
+
+
     # == intra-session modular functions == (requires a session and wont close it)
     def load_enzy_htp_stru(self, stru: Structure,
-                           pymol_session: pymol2.PyMOL) -> Tuple[str, pymol2.PyMOL]:
+                           session: pymol2.PyMOL) -> Tuple[str, pymol2.PyMOL]:
```
shaoqx, line 169: good catch on naming

enzy_htp/_interface/pymol_interface.py
```diff
@@ -8,77 +8,193 @@
...
-        argument.
+    def __init__(self, parent, config: PyMolConfig = None) -> None:
+        """Simplistic constructor that optionally takes an PyMOLConfig object as its only argument.
+        Calls parent class.
         """
-        self.config_ = config
-        if not self.config_:
-            self.config_ = default_pymol_config()
+        super().__init__(parent, config, default_pymol_config)
+        self.session_ = self.new_session()
```
shaoqx, line 41: I dont like the idea that PyMolInterface() object carries a session. Because the PyMolInterface object is singleton if it carries a session all the time the session will be super messed up. This does not fits the design of pymol2. A new sessions should be created everytime for a new independent task. (It is tested to be faster than using a session all the time in one of my previous code running for Lucy)
The session should be supplied as an argument **explicitly** to methods in the interface all the time. (trying to avoid the implicit IO)

shaoqx:
Thank you, Chris! Great changes! So many great new features!
In terms of changes:

- I really like the new design of a parent interface. It removes the bad "smell" of the code.
- I made some arguments about PyMolInterface and linting.
- One of the existing changes we want from develop_refactor in AmberInterface is changed back. Do you have any clue about why Git treats it like that?

enzy_htp/_interface/bcl_interface.py
chrisjurich: I disagree. The prefix `get_` just adds length to the function without adding a ton of meaning to the function.

enzy_htp/_interface/moe_interface.py
chrisjurich: I think it would be better to do something along the lines of `preparation.protonate.protonate_ligand(fname,method="MOE")`.

enzy_htp/core/typing.py
chrisjurich: Yes more or less. I found I was often writing code to check typing and I wanted to make the logging easier.

enzy_htp/mutation/engine.py
chrisjurich: Disregard this function for now. I will rewrite it and mimic what is being done for pymol/amber.

enzy_htp/structure/atom.py
chrisjurich: yeah i don't understand it either. I'll check out the existing version of the file from `develop_refactor`

enzy_htp/_interface/pymol_interface.py
chrisjurich: The reason it exists is that for all other `<Package>Interface`, an executable missing from the environment is not a problem until you try to run the function that uses it. So if the element isn't needed and is missing. it does not matter. `PyMolInterface()` is an exception since it relies on a python package. I wanted to preserve this behavior where the `enzy_htp` package can still run without having every element. So this is my solution.

enzy_htp/_interface/pymol_interface.py
chrisjurich: Sure, idrc.

enzy_htp/_interface/pymol_interface.py
chrisjurich: Noted and fair. I only use it for ligands but will add a warning about it.

enzy_htp/_interface/bcl_interface.py
shaoqx: My argument is, "get" makes it a verb phrase that the original noun phrase is not so intuitive as a function name (It sounds like formal_charge is a property of BCL unless you know what BCL is).
Also, it helps categorize methods in a class that in auto-complete, typing "get" will list you all the "get" type functions.

enzy_htp/_interface/moe_interface.py
shaoqx: The engine option is provided in `protonate_stru(stru, ph, engine, ligand_engine)`. And it will call `preparation.protonate.protonate_ligand_with_moe()` where there's additional handling from and to `enzy_htp.Structure` in addition to just "molfile" based IO. (the handling will be different for different engines. e.g.: pybel needs to fix the H naming)

enzy_htp/core/typing.py
shaoqx: cool

enzy_htp/mutation/engine.py
shaoqx: cool

enzy_htp/_interface/pymol_interface.py
shaoqx: Ohh I see, it exit(1), make sense.

enzy_htp/structure/atom.py
shaoqx: It seems a problem in old plum-dispatch version. I have it when my version is 1.5.9 in an environment and it goes away when I update it to 1.6 or to 2.1.1 using `pip`

enzy_htp/_interface/bcl_interface.py
chrisjurich: Ill meet you halfway. `get_` is an equally bad prefix. I'll change it to `calculate_formal_charge()`

enzy_htp/_interface/amber_interface.py
chrisjurich: Done with commit.

enzy_htp/_interface/amber_interface.py
chrisjurich: Done with commit.

enzy_htp/_interface/bcl_interface.py
chrisjurich: To my knowledge it only works for .sdf files. I have added that comment.

enzy_htp/_interface/bcl_interface.py
chrisjurich: Changed and committed.

enzy_htp/structure/atom.py
chrisjurich: fixed via a specific commit

test/mutation/mutation_pattern/test_position_pattern.py
chrisjurich: not sure tbh. I won't touch it for this commit but we can play around with the settings.

test/mutation/test_api.py
chrisjurich: See above comment. best way I can think for now is to disable yapf for this block.

test/mutation/mutation_pattern/test_target_aa_pattern.py
chrisjurich: See above. Lets figure it out in another PR

enzy_htp/_interface/pymol_interface.py
chrisjurich: This is specific to a problem I was having with the Rosetta paramterization script. I don't think it is applicable to the script you supplied. SDF's do not have atom names beyond element type and as a result there's a need to rename atoms from the BCL's conformer generation function.

enzy_htp/_interface/pymol_interface.py
chrisjurich: Made a commit about this

enzy_htp/_interface/pymol_interface.py
chrisjurich: Changed in a commit.

enzy_htp/_interface/pymol_interface.py
chrisjurich: thanks!

enzy_htp/_interface/pymol_interface.py
chrisjurich: Removed it.

test/mutation/test_mutation.py
chrisjurich: ^^

chrisjurich:
Have made all the necessary changes requested. In addition, I changed the API of `PyMolInterface` so that `session` is the first argument when present.

test/mutation/test_mutation.py
shaoqx: ^^

enzy_htp/_interface/bcl_interface.py
shaoqx: Oh I missed it

shaoqx (approved):
Great progress! This PR is now good to merge!

PR #116, opened 2023-07-10
Add support for PyMOL mutation engine

seb124:
This PR adds support for a PyMOL mutation engine through the definition of 2 new functions:

1.  Under mutation/api.py: `mutate_stru_with_pymol`. Takes in a Structure and List of Mutations. Creates a PyMOL session, loads the Structure into it, and calls the 2nd new function for each mutation in the List. Finally, saves the Structure from the PyMOL session and updates residues as necessary.

2. Under _interface/pymol_interface.py: `point_mutate`. Takes in mutation information and the current PyMOL object name and session. Performs a single point mutation in-place using cmd.wizard(). 

Also added tests for both of these functions under their respective test/ files. 

User will call this mutation engine with the mutation api's `mutate_stru` function and passing in "pymol" as the engine.


enzy_htp/mutation/api.py
```diff
@@ -367,8 +369,59 @@ def mutate_stru_with_tleap(
 
     return stru_cpy
 
+def mutate_stru_with_pymol(
+        stru: Structure,
+        mutant: List[Mutation],
+        in_place: bool = False,
+        ) -> Structure:
```
shaoqx, line 376: The path of the intermediate file could also be an special keyword arg to this function. (Reference int_leapout_pdb_path mutate_stru_with_tleap())

enzy_htp/mutation/api.py
```diff
@@ -367,8 +369,59 @@ def mutate_stru_with_tleap(
...
+    Args:
+        stru: the 'WT' structure
+        mutant: a list of Mutation() which describes a mutant to the 'WT'
+        in_place: if make the changes to the structure in-place 
+    Returns:
+        The mutated structure.
+    """
+
+    sp = PDBParser()
+    temp_file_name = fs.get_valid_temp_name("pymol_mutation_result.pdb")
```
shaoqx, line 387: Please reference the way temp files are handled in mutate_stru_with_tleap() so that they are by default under a system.SCRATCH_DIR shared across enzy_htp. (avoid messy workdir like old EnzyHTP did)

enzy_htp/mutation/api.py
```diff
@@ -367,8 +369,59 @@ def mutate_stru_with_tleap(
...
+        if not isinstance(mut, Mutation):
+            _LOGGER.error(
+                f"mutant takes only a list of Mutation(). Current mutant is: {mutant}")
+            raise TypeError
+
+    stru_cpy = copy.deepcopy(stru)
+
+    # 1. load stru into pymol
+    pi = interface.pymol
+    temp_session = pi.new_pymol_session()
```
shaoqx, line 400: The pymol_session is not closed. Please reference [this function](https://github.com/ChemBioHTP/EnzyHTP/blob/627bde52666d995680f5fd9f4580fc1e1e4179cf/enzy_htp/structure/structure_selection/general.py#L11) and use the context manager to handle the pymol session. (the design is like file IO)

enzy_htp/mutation/api.py
```diff
@@ -367,8 +369,59 @@ def mutate_stru_with_tleap(
...
+    pymol_obj_name, session = pi.load_enzy_htp_stru(stru_cpy, temp_session)
```
shaoqx, line 401: We dont need to have a new name for the `temp_session` here since load_enzy_htp_stru() just return the reference of the input temp_session. (probably not a good design lol)

enzy_htp/mutation/api.py
```diff
@@ -367,8 +369,59 @@ def mutate_stru_with_tleap(
...
+
+    # 2. loop through mutants and apply each one
+    for mut in mutant:
+        pi.point_mutate(mut.get_position_key(), mut.get_target(), pymol_obj_name, session)
+
+    # 3. save to a structure. TODO(SS): add function in pymol interface for this?
```
shaoqx, line 407: Yes, please make a general function in pymol interface for converting `objects in pymol sessions` to `enzy_htp.Structure`

enzy_htp/_interface/pymol_interface.py
```diff
@@ -101,6 +101,35 @@ def select_pymol_obj(self, pattern: str, pymol_obj_name: str,
...
+    def point_mutate(self, pos_key: Tuple[str, int], 
+                     target: str, pymol_obj_name: str, 
+                     pymol_session: pymol2.PyMOL) -> None:
+        """
+        Performs a single point mutation on the PDB file in the PyMOL session in-place.
+        Args:
+            pos_key: the chain id and residue index.
+            target: the target residue name (3 letters).
+            pymol_obj_name: the name of the PDB file in PyMOL.
+            pymol_session: the current PyMOL session.
```
shaoqx, line 113: current -> target

the point of pymol2 is you can have many pymol session working at the same time so current dont make much sense

enzy_htp/_interface/pymol_interface.py
```diff
@@ -101,6 +101,35 @@ def select_pymol_obj(self, pattern: str, pymol_obj_name: str,
...
 
+    def point_mutate(self, pos_key: Tuple[str, int], 
+                     target: str, pymol_obj_name: str, 
+                     pymol_session: pymol2.PyMOL) -> None:
+        """
+        Performs a single point mutation on the PDB file in the PyMOL session in-place.
+        Args:
+            pos_key: the chain id and residue index.
+            target: the target residue name (3 letters).
+            pymol_obj_name: the name of the PDB file in PyMOL.
```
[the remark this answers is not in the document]
shaoqx: And the same for the description above in the docstring

enzy_htp/_interface/pymol_interface.py
```diff
@@ -101,6 +101,35 @@ def select_pymol_obj(self, pattern: str, pymol_obj_name: str,
...
+        """
+
+        # load pymol wizard mutagenesis function
+        pymol_session.cmd.wizard("mutagenesis")
+        pymol_session.cmd.do("refresh_wizard")
+        
+        # select res idx to mutate to target
+        pymol_session.cmd.get_wizard().set_mode(target)
+        if pymol_obj_name == "":
+            pymol_session.cmd.get_wizard().do_select(pos_key[0] + "/" + str(pos_key[1]) + "/")
```
shaoqx, line 125: I suggest usingh f-string when handling string like this
https://realpython.com/python-f-strings/

shaoqx:
Great job! I added some quick comments.

enzy_htp/_interface/pymol_interface.py
```diff
@@ -101,6 +101,35 @@ def select_pymol_obj(self, pattern: str, pymol_obj_name: str,
...
+        else:
+            pymol_session.cmd.get_wizard().do_select("/" + pymol_obj_name + "//" + pos_key[0] + "/" + str(pos_key[1]) + "/")
+
+        # select the best rotamer (defaulted to frame 1) and apply
+        pymol_session.cmd.frame(1)
```
shaoqx, line 130: IMPORTANT:
It seems the lowest conformer is not frame 1 lol.
I tested using the unitest you wrote and I compared the structure and found frame 1 is just frame 1 and the best one should be frame 2 in this case. So we prob still need to determine it. (Maybe try to find something like current frame. or try just apply instead of choosing the frame. the most ideal will be obtaining the complete information from pymol)

test/_interface/test_pymol_interface.py
```diff
@@ -67,3 +67,16 @@ def test_select_pymol_obj():
...
+
+def test_point_mutate():
+    """testing that point mutations work as expected"""
+    test_stru = PDBParser().get_structure(f"{DATA_DIR}KE_07_R7_2_S.pdb")
+    test_mut_stru = PDBParser().get_structure(f"{DATA_DIR}KE_07_R7_2_S_mut.pdb")
+    pi = interface.pymol
+    test_session = pi.new_pymol_session()
+    pymol_obj_name, session = pi.load_enzy_htp_stru(test_stru, test_session)
+    pi.point_mutate(("A", 154), "TRP", pymol_obj_name, session)
+    assert len(pi.select_pymol_obj("resi 154", pymol_obj_name, session)) == (
```
shaoqx, line 80: I found the reason why the atom index are different. Pymol add all the new atoms at the end. You can try the `sort` command in pymol about this. (It sorts when saving by default tho)

enzy_htp/mutation/api.py
shaoqx: `retain_order` seems not necessary. It will make the saved PDB looks weird. You can try look into the saved PDB.

shaoqx: It is pretty cool how enzy_htp can handle the weird looking PDB though. I looked into the get_structure again and indeed the algorithm is general enough to handle this.

enzy_htp/_interface/pymol_interface.py
shaoqx: Green is WT. Blue is from enzy_htp using this function. White is from current pymol session.
Best frame:
image
Frame 1:
image

test/_interface/test_pymol_interface.py
shaoqx: image

enzy_htp/mutation/api.py
shaoqx: image
This is what retain_order enabled saving looks like

enzy_htp/_interface/pymol_interface.py
seb124: image

It seems as though the PyMOL in python has a different rotamer library, as it looks like the mutagenesis is picking frame 6 from 9 rather than frame 2 from 8. This is without cmd.frame(1). Green is WT, Blue is using enzy_htp, white is current session.

Frame 1:
image

Frame 2 (best frame):
image

It doesn't seem like it's explicitly choosing frame 1 anymore, which is a plus.

enzy_htp/mutation/api.py
seb124: When I remove retain_order, it looks much better:

image

However, the test function fails as the atoms are no longer in the same order as they were before. Is this okay? I can change the test if so.

test/_interface/test_pymol_interface.py
seb124: image

Even with sort, it looks like the atom indices are still off. Since it sorts when saving by default, though, should I keep the test the same for now?

enzy_htp/_interface/pymol_interface.py
shaoqx: I checked out the MD for this mutant and it is indeed "frame 2" in GUI is the most frequent conformer.

shaoqx: This could be a reference https://pymolwiki.org/index.php/Rotkit

seb124: Looks like PyMOL (and Rotkit) considers frame 1 as the most probable mutation based off the percentage shown below:

image

The "31.9%" next to "mutation" is the probability of the mutation actually occurring with this rotamer. The percentages are listed in decreasing order as the frames go on, which is why it defaults to frame "1" I think.

enzy_htp/mutation/api.py
shaoqx: It really is! Good catch on the atom order. We will keep "retain_order" then.
Maybe add another test with a multichain structure where a mutation is deployed on chain B. In principle, it also works.

enzy_htp/_interface/pymol_interface.py
seb124: image

So I've found a way to print them all out. It seems like in this library, rotamer 6 (the one defaulted to when removing "frame(1)") has the lowest strain, leading me to believe the mutagenesis wizard defaults to the lowest strain mutation and applies it. If this is true, no more needs to be done, although the more common rotamer you found using MD won't be chosen due to the fact that the rotamer library isn't fully updated with the GUI for whatever reason.

shaoqx: Wow great job!
I noticed the percentages too. I searched about it but didn't find any explanation. Frame 1 clearly have multiple clashes with a super high strain score comparing to others. Could it be the percentage of the equivalent rotamers? (i.e.: the permutation of all possible dihedrals have most of them corresponding to this rotamer)

shaoqx: Great job finding the way to print out the scores again!
It will be great if we can have an argument to print those scores out and save the corresponding structures for debugging purposes.
For doing mutation, we can keep the current strategy of just removing frame(1).
For the inconsistency between pymol2 and pymol-GUI, we should open an issue in their [GitHub repo](https://github.com/schrodinger/pymol-open-source/issues). This is a good chance to practice writing an issue on GitHub. The key point is not to involve too much about EnzyHTP, isolate the problem itself and report to them.

shaoqx:
I notice a problem related to the atom naming. I made it an issue since it's prob better to do it after this PR. #117 You can assign yourself to the issue and solve it after this PR.

enzy_htp/mutation/api.py
seb124: Keeping "retain_order" on a multichain structure doesn't seem like it will work:

image

To solve this, I was going to just check the # of chains and turn "retain_order" off if it was >1, but I found another problem that I think has to do with how pymol2 saves its PDBs. Whenever it saves a structure with HETATMs and CONECTs, it gets rid of the CONECTs and changes the HETATMs to ATOMs. This causes the following error when enzy_htp tries to load it into a structure:

image

When I save a structure with the GUI, this problem does not occur, as it keeps the HETATMs and CONECTs. Not sure if there's a cmd/param in cmd.save to fix this, but I am currently looking.

enzy_htp/_interface/pymol_interface.py
seb124: Ok, I just added a "debug" parameter to point_mutate() that does this

enzy_htp/mutation/api.py
shaoqx: Omg the behavior on multichain is ridiculous. My suggestion is get rid of retain_order for all cases and in your test, you can assert atoms in the same residue using set() to make them order less.
We will write another function supporting maximumly restoring the atom order together with the name fix function.

shaoqx: Also if you dont use retain_order, will the error below still pop off?

seb124: Yeah, it'll still give the same error, even if I don't mutate the WT at all

shaoqx: But I think in a EnzyHTP saved PDB there should be nothing but ATOM records. Can you send me the PDB file from Pymol that causes this error?

seb124: It's puo_put.pdb in the test/mutation/data folder

enzy_htp/_interface/pymol_interface.py
```diff
@@ -101,6 +101,86 @@ def select_pymol_obj(self, pattern: str, pymol_obj_name: str,
 
         return sorted(result)
 
+    def point_mutate(self, pos_key: Tuple[str, int],
```
shaoqx, line 104: export_enzy_htp_stru() could be a better name so that it is symmetrical with load_enzy_htp_stru()

enzy_htp/_interface/pymol_interface.py
```diff
@@ -101,6 +101,86 @@ def select_pymol_obj(self, pattern: str, pymol_obj_name: str,
...
+                pymol_session.cmd.get_wizard().do_state(i)
+                pymol_session.cmd.frame(i)
+                pymol_session.cmd.get_wizard().apply()
+                self.save_to_stru(pymol_obj_name, pymol_session, True)
+            return
+
+        pymol_session.cmd.get_wizard().apply()
+
+
+    def save_to_stru(self, pymol_obj_name: str,
```
shaoqx, line 152: export_enzy_htp_stru() could be a better name so that it is symmetrical with load_enzy_htp_stru()

enzy_htp/_interface/pymol_interface.py
```diff
@@ -101,6 +101,86 @@ def select_pymol_obj(self, pattern: str, pymol_obj_name: str,
...
+        temp_dir = eh_config["system.SCRATCH_DIR"]
+        fs.safe_mkdir(temp_dir)
+        pymol_outfile_path = fs.get_valid_temp_name(
+                f"{eh_config['system.SCRATCH_DIR']}/pymol_output.pdb")
+
+        if in_order:
+            pymol_session.cmd.set("retain_order")
+
+        pymol_session.cmd.select(pymol_obj_name)
+        pymol_session.cmd.save(pymol_outfile_path, "(sele)")
```
shaoqx, line 176: Would it be different from just use the pymol_obj_name instead of (sele) and avoid the redundant cmd.select?

enzy_htp/_interface/pymol_interface.py
```diff
@@ -101,6 +101,86 @@ def select_pymol_obj(self, pattern: str, pymol_obj_name: str,
...
+                pymol_session.cmd.get_wizard().apply()
+                self.save_to_stru(pymol_obj_name, pymol_session, True)
+            return
+
+        pymol_session.cmd.get_wizard().apply()
+
+
+    def save_to_stru(self, pymol_obj_name: str,
+                    pymol_session: pymol2.PyMOL,
+                    in_order: bool = True,
```
shaoqx, line 154: I suggest we keep the original naming that exists as much as possible so that it is easier to understand just looking at the name. In other word, I suggest name it "if_retain_order". (the if_xxx is kind of my thing and it is all over enzy_htp)

enzy_htp/_interface/pymol_interface.py
```diff
@@ -101,6 +101,86 @@ def select_pymol_obj(self, pattern: str, pymol_obj_name: str,
...
+            in_order: if the saving should keep the order of the atoms in the original object.
+            keep_files: if the program should clean up the temp files or not.
+        Returns:
+            A Structure object representing the target enzyme in PyMOL.
+        """
+        sp = PDBParser()
+        temp_dir = eh_config["system.SCRATCH_DIR"]
+        fs.safe_mkdir(temp_dir)
+        pymol_outfile_path = fs.get_valid_temp_name(
+                f"{eh_config['system.SCRATCH_DIR']}/pymol_output.pdb")
```
shaoqx, line 170: You already have temp_dir assigned so you can just use it here for code clarity.

enzy_htp/_interface/pymol_interface.py
```diff
@@ -101,6 +101,86 @@ def select_pymol_obj(self, pattern: str, pymol_obj_name: str,
...
+                pymol_session.cmd.get_wizard().set_mode(target)
+                if pymol_obj_name == "":
+                    pymol_session.cmd.get_wizard().do_select(f"{pos_key[0]}/{str(pos_key[1])}/")
+                else:
+                    pymol_session.cmd.get_wizard().do_select(f"{pymol_obj_name}//{pos_key[0]}/{str(pos_key[1])}/")
+            
+                pymol_session.cmd.get_wizard().do_state(i)
+                pymol_session.cmd.frame(i)
+                pymol_session.cmd.get_wizard().apply()
+                self.save_to_stru(pymol_obj_name, pymol_session, True)
```
shaoqx, line 146: I think this dont really save the file right? The "True" is given to "in_order" instead of "keep_file"

shaoqx:
Great job resolving those comments! I added some more comments to the new changes you made.
(I tried to make them very details for the training purpose.)

enzy_htp/_interface/pymol_interface.py
shaoqx: Good job! Please let me know if you opened an issue for open-source-pymol and I will close this comment after that.

enzy_htp/mutation/api.py
shaoqx: Ohhhh I found the reason. We were not allowing multiple chain id to show up in the ATOM chain.
(The error information is wrong. My mistake.) Would you allow me to commit a change fixing this in your branch?

shaoqx: I think a good plan is to add an option in get_structure about whether allowing it. It does not follow the defination of PDB but we can turn it on for rare cases like this

seb124: Yeah that sounds good, you can commit the change

enzy_htp/_interface/pymol_interface.py
seb124: I just opened an issue!

enzy_htp/mutation/api.py
shaoqx: Cool! I commit some changes that fixed the problem by enabling allow_multichain_in_atom=True (find reference in the test I added)
You can use this option for your export_to_pdb method

enzy_htp/_interface/pymol_interface.py
```diff
@@ -101,6 +102,111 @@ def select_pymol_obj(self, pattern: str, pymol_obj_name: str,
...
+            
+                pymol_session.cmd.get_wizard().do_state(i)
+                pymol_session.cmd.frame(i)
+                pymol_session.cmd.get_wizard().apply()
+                self.export_pymol_obj(pymol_obj_name, pymol_session, rotamer=i)
+            return
+
+        pymol_session.cmd.get_wizard().apply()
+
+    def export_pymol_obj(self, pymol_obj_name: str,
```
shaoqx, line 152: a better name aligning with export_enzy_htp_stru could be export_pdb (since we are always exporting pymol obj but to different target formats)
I didn't suggest export_file because we may add some treatment that specific to PDB (e.g.: whether use retain order)
Maybe in some day we need more than just PDB (e.g.: mol2) we will consider making a very general one export_file and call it in export_pdb

enzy_htp/_interface/pymol_interface.py
```diff
@@ -101,6 +102,111 @@ def select_pymol_obj(self, pattern: str, pymol_obj_name: str,
...
+                pymol_session.cmd.get_wizard().apply()
+                self.export_pymol_obj(pymol_obj_name, pymol_session, rotamer=i)
+            return
+
+        pymol_session.cmd.get_wizard().apply()
+
+    def export_pymol_obj(self, pymol_obj_name: str, 
+                        pymol_session: pymol2.PyMOL,
+                        if_retain_order: bool = True,
+                        rotamer: int = 0) -> str:
```
shaoqx, line 155: Since this function is very general to saving PDB. It should know nothing about doing mutations. An alternative design that decouples its dependence on mutation knowledge is to change "rotamer" to "tag" or "name_tag" and in point_mutate() we figure out what name tag is passed to this function. (it could be something like "_point_mutate_R154W_rot_1")

enzy_htp/_interface/pymol_interface.py
```diff
@@ -101,6 +102,111 @@ def select_pymol_obj(self, pattern: str, pymol_obj_name: str,
...
+        sp = PDBParser()
+        temp_dir = eh_config["system.SCRATCH_DIR"]
+        fs.safe_mkdir(temp_dir)
+        pymol_outfile_path = fs.get_valid_temp_name(
+                f"{temp_dir}/pymol_output.pdb")
+
+        if if_retain_order:
+            pymol_session.cmd.set("retain_order")
+
+        pymol_session.cmd.save(pymol_outfile_path, pymol_obj_name)
```
shaoqx, line 204: We can use the newly made export_pdb to avoid repeating the same code. (sorry if I commented too early that you were about to do this)

enzy_htp/_interface/pymol_interface.py
```diff
@@ -101,6 +102,111 @@ def select_pymol_obj(self, pattern: str, pymol_obj_name: str,
...
+        res = sp.get_structure(pymol_outfile_path)
```
shaoqx, line 205: Please try the new option I made and test it with a multichain system

shaoqx:
Good job resolving the comments!
I made some final suggestions and I think we will be ready to merge after resolving these!

enzy_htp/_interface/pymol_interface.py
```diff
@@ -101,6 +102,104 @@ def select_pymol_obj(self, pattern: str, pymol_obj_name: str,
...
+            pymol_obj_name: the name of the target enzyme in PyMOL.
+            pymol_session: the target PyMOL session.
+            in_order: if the saving should keep the order of the atoms in the original object.
+            if_multichain: if the Structure object created will have more than one chain.
+        Returns:
+            A Structure object representing the target enzyme in PyMOL.
+        """
+        sp = PDBParser()
+        pymol_outfile_path = self.export_pdb(pymol_obj_name, pymol_session, if_retain_order=if_retain_order)
+        res = sp.get_structure(pymol_outfile_path, allow_multichain_in_atom=if_multichain)
```
shaoqx, line 198: I think this option can always be True in this function. Since the caller of this function will not necessarily know if the object have multichain. Let just hard code it as True here and remove the if_multichain and change when problems show up.

enzy_htp/mutation/api.py
```diff
@@ -367,8 +371,62 @@ def mutate_stru_with_tleap(
...
+
+    if_retain_order = True
+
+    # if structure is multichain, set retain_order to false and raise a flag to mark multichain structure
+    # for further treatment
+    if stru_cpy.num_chains > 2:
+        if_multichain = True
+        if_retain_order = False
+    else:
+        if_multichain = False
```
shaoqx, line 405: let's just always don't use retain_order here and address the mess-up atom order in next PR.

shaoqx:
Cool! Just several last changes and confirm the unit tests and we can merge it!

shaoqx (approved):
Great efforts implementing PyMol as the mutation engine!
We will address the atom naming and order issue brought by pymol <-> PDB in the next PR.

Issue #117, opened 2023-07-12
Atom naming consistency


shaoqx:
Another thing I noticed is the atom names of the new residue generated by PyMol are not the classical PDB style naming. I compared and found it is mainly the hydrogen namings are different. Fixing this is a good lv. 3 task! I suggest we can choose from the 2 plans below:
1. Assume H naming is the only difference:
         detect the inconsistency between atom namings after the mutation. (you can use the standard atom naming information from chemical.residue) and change all the naming of H back (it's just simply where to place the digit) and abort when it is more than just H
2. Difference in naming in general:
         still detect the inconsistency between atom namings after the mutation. (this should be a general method of Structure or Atom)
         Wrote some 1 time code and collect a naming map between our standard atom naming and pymol's atom naming. (this can be achieved by make pymol generate each residue and record their atom naming.) And store it in chemical.

shaoqx:
Atom names are used in many places in EnzyHTP as the key to some value. We decided to use the one in chemical.residue as the standard one for EnzyHTP and convert any that does not conform.
See the idea in https://github.com/ChemBioHTP/EnzyHTP/blob/627bde52666d995680f5fd9f4580fc1e1e4179cf/enzy_htp/chemical/metal.py#L155

shaoqx:
Here is an example of PyMol atom names for TRP comparing to native ones (sorted by A-Z)
![image](https://github.com/ChemBioHTP/EnzyHTP/assets/36098720/5f755c37-eb69-4664-ab5f-b51dccff709b)

shaoqx:
Test tleap behavior on it

Issue #118, opened 2023-07-13
Fix the messed up atom order from PyMol


shaoqx:
`retain_order` seems not necessary. It will make the saved PDB looks weird. You can try look into the saved PDB.

_Originally posted by @shaoqx in https://github.com/ChemBioHTP/EnzyHTP/pull/116#discussion_r1259127656_

shaoqx:
It causes error in the multichain system. We should not use retain_order and create a function to restore the order.

shaoqx:
A good design would be having an `align_atom_order(stru, ref_stru)` function in `structure_operation.general` as this function stands for aligning atom orders/indexes of Structure()s in general. Because in this issue, the method of fixing is independent of pymol knowledge. And we call it after` export_enzy_htp_stru()` in api.py

PR #120, opened 2023-07-26
Fix PyMOL Saving Issues

seb124:
Fixes issues #117 and #118 by adding 2 functions under **structure.py**:

`fix_pymol_naming`: loops through every Atom in the Structure and updates names of atoms if they are in the `PYMOL_TO_STRU` dictionary, which was added under **chemical/residue.py.** 

`order_to_stru`: takes two Structures (`stru` and `ordered_stru`). Compares residues between both, and if they are the same, updates the order of `stru` to match `ordered_stru`.

Also added/updated tests for both of these functions and implemented the two functions into `export_pdb` and `export_enzy_htp_stru`, respectively. Now, when a Structure is saved from PyMOL, it will have the same order and naming as the original Structure passed in (excluding mutated regions, where only the naming will be consistent with that of the Structure class).


enzy_htp/structure/structure.py
```diff
@@ -964,6 +965,22 @@ def build_metalcenters(self, dir, ft="PDB"):
...
+        """
+        for chain in self.chains:
+            for residue in chain.residues:
+                for atom in residue.atoms:
+                    # special case; HA in GLY becomes HA2
+                    if residue.name == "GLY" and atom.name == "HA":
+                        atom.name = "HA2"
+
+                    if atom.name in PYMOL_TO_ATOM_MAPPER:
+                        atom.name = PYMOL_TO_ATOM_MAPPER[atom.name]
```
shaoqx, line 984: This is not a good design of this function. The Structure class should never know anything about PyMol. Every knowledge EnzyHTP knows about PyMol should go to PyMolInterface and pymol_config. 
Also this function here can be decomposed into 2 more modular functions:
1. give the atom mapping that pymol fix needs.
2. batch edit atom names in Structure() given a atom naming mapper as a input. (You can also make sth like this to match the GLY behavior:
{"GLY":{"HA" : "HA2"}, "DEFAULT":default_mapper} and make a getter function for the dict like we did in /chemical
)
The 2nd function can stay in Structure class

enzy_htp/structure/structure.py
```diff
@@ -964,6 +965,22 @@ def build_metalcenters(self, dir, ft="PDB"):
...
+            stru: the Structure that contains the wrong atom names (if any).
+        Returns:
+            Nothing.
+        """
+        for chain in self.chains:
+            for residue in chain.residues:
+                for atom in residue.atoms:
+                    # special case; HA in GLY becomes HA2
+                    if residue.name == "GLY" and atom.name == "HA":
+                        atom.name = "HA2"
```
shaoqx, line 980: Make sure the chirality is not messed up. Is HA2 always the [pro-S](https://chem.libretexts.org/Bookshelves/Organic_Chemistry/Book%3A_Organic_Chemistry_with_a_Biological_Emphasis_v2.0_(Soderberg)/03%3A_Conformations_and_Stereochemistry/3.12%3A_Prochirality) hydrogen.

enzy_htp/_interface/pymol_interface.py
```diff
@@ -279,12 +279,22 @@ def export_pdb(self, pymol_session: pymol2.PyMOL, pymol_obj_name: str, if_retain
             pymol_session.cmd.set("retain_order")
         pymol_session.cmd.save(pymol_outfile_path, pymol_obj_name)
 
+        # fix the atom naming issue.
+        temp_stru = PDBParser().get_structure(pymol_outfile_path,allow_multichain_in_atom=True)
+        temp_stru.fix_pymol_naming()
```
shaoqx, line 284: Naming should not be fixed here since the naming error is not caused by the general file saving behavior of PyMol but by those pymol functions that generates new structures. (e.g.: point mutation)
But this design is not easy since the point_mutate function do not have access to the naming change (or at least we need to write new interface for it). I can suggest to new design here:
1. move it to export_enzy_htp_stru() and turning it off by default (manually turn it on when calling from the mutation api)
2. change the fix_pymol_naming by changing the atom names using pymol's `alter` and call it in point_mutate

enzy_htp/_interface/pymol_interface.py
```diff
@@ -297,6 +307,11 @@ def export_enzy_htp_stru(self,
         sp = PDBParser()
         pymol_outfile_path = self.export_pdb(pymol_obj_name, pymol_session, if_retain_order=if_retain_order)
         res = sp.get_structure(pymol_outfile_path, allow_multichain_in_atom=True)
+
+        # update order of non-mutated residues
+        if ordered_stru is not None:
+            order_to_stru(res, ordered_stru)
```
shaoqx, line 313: Sorry I kinda presented it not very clear in the issue that this functions should be called "after" export_enzy_htp_stru() is called in api.py, which means it should be in mutate_stru_with_pymol()

enzy_htp/chemical/residue.py
```diff
@@ -560,6 +560,45 @@
 
 # yapf: enable
 
+"""A map to convert messed up PyMOL atom names to a more conventional name for the Structure class."""
+PYMOL_TO_ATOM_MAPPER: Dict[str, str] = {
```
shaoqx, line 564: pymol_config is the place for this mapper

shaoqx:
Great job fixing the 2 issues! I pointed out some design that we can improve.

enzy_htp/_config/pymol_config.py
```diff
@@ -18,6 +18,44 @@
...
+        "2HG1": "HG12",
+        "3HG1": "HG13",
+        "1HD1": "HD11",
+        "2HD1": "HD12",
+        "3HD1": "HD13",
+        "1HD2": "HD21",
+        "2HD2": "HD22",
+        "3HD2": "HD23",
+    }
+
```
shaoqx, line 58: The docstring should go under the object defination

enzy_htp/_interface/pymol_interface.py
```diff
@@ -278,27 +278,50 @@ def export_pdb(self, pymol_session: pymol2.PyMOL, pymol_obj_name: str, if_retain
...
+                            pymol_session: pymol2.PyMOL,
+                            if_retain_order: bool = False,
+                            if_fix_naming: bool = False) -> Structure:
         """
         Saves a PyMOL object to a Structure object.
         Args:
             pymol_session: the target PyMOL session.
             pymol_obj_name: the name of the target enzyme in PyMOL.
             if_retain_order: if the saving should keep the order of the atoms in the original object.
+            if_fix_naming: if the PyMOL naming of atoms should be updated to match the Structure convention.
```
shaoqx, line 295: Also state the problem for both of them. Giving a link pointing to the two issue each is a good idea

enzy_htp/structure/structure.py
```diff
@@ -962,9 +962,20 @@ def build_metalcenters(self, dir, ft="PDB"):
...
+    #     Args:
+    #         map: the map of incorrect atom names to correct atom names.
+    #     Returns:
+    #         Nothing.
+    #     """
+    #     for chain in self.chains:
+    #         for residue in chain.residues:
+    #             for atom in residue.atoms:
+    #                 if atom.name in atom_map:
+    #                     atom.name = atom_map[atom.name]
```
shaoqx, line 976: Just delete this is fine

enzy_htp/structure/structure.py
```diff
@@ -1004,3 +1015,16 @@ def merge_right(left: Structure,
...
+    """Orders the atoms in stru in-place to match the ordered_stru's ordering.
+    Args:
+        stru: 
+        ordered_stru: the Structure object that has the desired order.
+    Returns:
+        Nothing.
+    """
+    for res, ordered_res in zip(stru.residues, ordered_stru.residues):
+        if res.name == ordered_res.name:
+            res.atoms = ordered_res.atoms
```
shaoqx, line 1030: This is problematic.

shaoqx: This function should go to `structure_operation/general.py`. Sorry for the misleading direction before.

shaoqx: It will be great to rename ordered_stru as ref_stru and change the name of the function more specific. By specific, I mean this function cannot handle cases like stru and ref_stru dont have the same order/number of residues. Such limitation should be reflected on the function name.

enzy_htp/_config/pymol_config.py
```diff
@@ -57,7 +95,20 @@ def required_executables(self) -> List[str]:
     def required_py_modules(self) -> List[str]:
         """ """
         return ["pymol2"]
+    
+    def update_atom_name(self, res: str, atom: str) -> str:
```
shaoqx, line 99: "update" is pretty vague. Could get_canonical_atom_name a better name?

enzy_htp/structure/structure.py
shaoqx: For the new algorithm, I assume that you forgot to delete `res.atoms = ref_res.atoms`
But still, there could be a better way of reordering a list. We can discuss it in our meeting tomorrow.

test/mutation/test_api.py
```diff
@@ -157,7 +157,8 @@ def test_mutate_stru_with_pymol():
             assert new_res.name == "TRP"
             assert len(new_res.atoms) == 24
         else:
-            assert len(new_res.atoms) == len(old_res.atoms)
+            for new_atom, old_atom in zip(new_res.atoms, old_res.atoms):
+                assert new_atom.name == old_atom.name
 
     # test with multiple mutants
```
shaoqx, line 163: Let's make them seperate test functions so that we know the type when one fails

enzy_htp/structure/structure_operation/general.py
```diff
@@ -93,3 +93,20 @@ def update_residues(resi: Residue, ref_resi: Residue) -> Residue:  # pylint: dis
         resi.name = ref_resi.name
     resi.atoms = copy.deepcopy(ref_resi.atoms)  # this will also set resi as parent
     return resi
+
+
+def align_atom_order_in_each_residue(stru: Structure, ref_stru: Structure) -> Structure:
```
shaoqx, line 98: we need a test for this function

shaoqx (approved):
Great job addressing all the comments! This is a fantastic fix! We also fixed an important bug about Structure deepcopy!

PR #128, opened 2023-09-20
Reactive Docking PR 1

chrisjurich:
This is a fairly large PR that includes code needed to perform Reactive Docking in EnzyHTP. There are a few major categories of changes implemented in this branch:

1. Config behavioral changes. 
+ I have added a new abstract class called `BaseConfig()` which all Config classes should inherit from. It enables uniform getting/setting with `[]` operators. This file is found in `enzy_htp/_config/base_config.py` and all `<Package>Config()` classes in `enzy_htp/_config` have been updated to inherit from this. 
+ I have added the ability of `enzy_htp` to look for a global config file at `$HOME/.eh_config` at start up. This enables easy re-use of configuration settings. Code supporting this functionality is found in `enzy_htp/_config/__init__.py`. 
2. New Software Interfaces
+ I have added access for AlphaFill, RDKit, and xtb. All corresponding `<Package>Interface()` and `<Package>Config()` classes are found in `enzy_htp/_interface/` and `enzy_htp/_config`, respectively. 
3. New analytical methods
+ A function which approximates binding energy can be found in `enzy_htp/analysis/binding_energy.py` 
4. Reactive docking functionality
+ Reactive docking functionality has been added to the `enzy_htp.preparation` module. It is accessed via the `dock_reactants()` function and makes use of many of the new functionalities described elsewhere in this PR
5. Updated existing software interfaces
+ Made mutations with RosettaScripts possible
+ Created functionality for protein backbone capping in pymol.


enzy_htp/_config/__init__.py
```diff
@@ -7,10 +7,31 @@
...
+if fs.has_content(config_file):
+    _LOGGER.info(f"Found config file: {config_file}...")
+    lines: List[str] = fs.lines_from_file(config_file)
+    counter: int = 0
+    for ll in lines:
+        tks = ll.split('#')
+        tk = tks[0].strip()
+        if not tk:
+            continue
+        exec(tk)
```
shaoqx, line 34: The use of exec is quite dangerous. It also makes the syntax of eh_config out of control. Please mark a TODO for changing it in the future.

enzy_htp/_config/mole2_config.py
```diff
@@ -0,0 +1,57 @@
...
+"""
+
+from typing import List, Any
+
+from copy import deepcopy
+
+from .base_config import BaseConfig
+
+
+class Mole2Config(BaseConfig):
```
shaoqx, line 16: fun!

enzy_htp/_interface/base_interface.py
```diff
@@ -20,7 +19,7 @@
 from enzy_htp.core import env_manager as em
 
 
-class BaseInterface:
+class BaseInterface(ABC):
```
shaoqx, line 22: How does it work when you have an ABC without any @abstractmethod

shaoqx: I see. It is a blue print that is not strictly enforced

enzy_htp/_interface/pymol_interface.py
```diff
@@ -519,6 +522,113 @@ def get_sequence(self, session: pymol2.PyMOL, fname: str, sele: str = '(all)') -
...
+
+            self.general_cmd(session, args)
+
+            args = [('valence', 'guess', 'name C21 or name C22'), ('h_add', 'name C21'), ('h_add', 'name C22'), ("save", outfile, obj_name)]
+            self.general_cmd(session, args)
+
+        return outfile
+
+
+    def center_of_mass(self, session, sele:str='all', no_hydrogens:bool=True):
```
shaoqx, line 591: This can be realized by Structure() inside EnzyHTP btw

enzy_htp/mutation/api.py
```diff
@@ -406,6 +410,102 @@ def mutate_stru_with_pymol(
...
+        stru_oper.update_residues(stru_cpy, rosetta_stru)
+        return stru_cpy
+    #TODO(CJ): do it this way
+    stru_oper.update_residues(stru_cpy, pymol_mutant_stru)
+
+    if in_place:
+        stru_oper.update_residues(stru, stru_cpy)
+        return stru
+
+    return stru_cpy
```
shaoqx, line 506: Maybe clean this part for the PR?

enzy_htp/mutation/api.py
```diff
@@ -406,6 +410,102 @@ def mutate_stru_with_pymol(
...
+
+    #stru_cpy = copy.deepcopy(stru)
+
+    if_retain_order = True
+    #TODO(CJ): need to remap the residue names at some point
+    #TODO(CJ): use the scratch_dir
+    parser = PDBParser()
+    stru_content: str = parser.get_file_str(stru)
+    temp_file: str = '__mut_temp.pdb'
+    fs.write_lines(temp_file, stru_content.splitlines())
```
shaoqx, line 465: We can wrap this part as a method in RosettaInterface that convert Structure() -> path_of_rosetta_happy_pdb. As far as I know Rosetta dont like names like HID HIP etc. and need a {rosetta_home}/tools/protein_tools/scripts/clean_pdb.py process in general according to Rocco.

enzy_htp/preparation/align_ligand.py
```diff
@@ -0,0 +1,196 @@
...
+import rdkit.Chem as _rchem
+from rdkit.Chem import rdMolTransforms as rdmt
+
+from enzy_htp import config, interface
+
+from enzy_htp.core import file_system as fs
+from enzy_htp.core import _LOGGER
+
+
+def align_ligand(template: str,
```
shaoqx, line 24: We should have most thing outside of the _interface take and output objects in /strcuture.
In this way the protocol is not coupled with a specific file type. 
For this function, it will be great if they IOs are Ligand()s

enzy_htp/preparation/align_ligand.py
```diff
@@ -0,0 +1,196 @@
...
+"""
+
+from collections import defaultdict
+from typing import List, Dict, Set, Tuple
+from pathlib import Path
+
+import pandas as pd
+import numpy as np
+import rdkit.Chem as _rchem
+from rdkit.Chem import rdMolTransforms as rdmt
```
shaoqx, line 16: Can the use of these go to RDKitInterface as well as the data structure conversion (Structure() -> Molecule())?

enzy_htp/preparation/align_ligand.py
shaoqx: Looking more into this make me feel the current version of the function should go to RDKit interface and leave just an engining use "science API" here

enzy_htp/preparation/cluster.py
```diff
@@ -0,0 +1,77 @@
...
+import pandas as pd
+
+from enzy_htp import interface, config
+
+import enzy_htp.chemical as chem
+
+from enzy_htp.core import file_system as fs
+
+
+def create_cluster(fname: str, sele_str: str, outfile: str = None, cap_strategy: str = 'H', work_dir: str = None, session=None) -> str:
```
shaoqx, line 13: Same as comments in align_ligand.py.
- /structure objs as IO (you can refer to the use in mutation module)

shaoqx: Oh I saw you have it in the PyMolInterface. Which version is the final one? Is this function waiting for update?

enzy_htp/preparation/reactive_docking.py
```diff
@@ -0,0 +1,1285 @@
...
+                   rng_seed: int = 1996,
+                   n_struct: int = 1000,
+                   cst_cutoff: int = 10,
+                   clash_distance: float = 2.0,
+                   clash_cutoff: int = 1,
+                   binding_cutoff: int = 20,
+                   cluster_distance: float = 2.5,
+                   use_cache: bool = True,
+                   parent: bool = False,
+                   pdb_code: None = str) -> List[str]:
```
shaoqx, line 44: It will be great if the output can be a Structure() or list of them

enzy_htp/preparation/constrained_docking.py
```diff
@@ -0,0 +1,564 @@
...
+
+def constrained_docking(structure: Structure,
+                        reactants: List[str],
+                        param_files: List[str],
+                        constraints: List[RosettaCst],
+                        work_dir: str = None,
+                        use_cache: bool = False,
+                        rng_seed: int = 1996,
+                        n_struct: int = 100,
+                        minimize: bool = True) -> pd.DataFrame:
```
shaoqx, line 33: Now I get the idea of the complete thing is not finished rn so the IOs are not finalize.

Just make sure in the final version, functions the can be use not just for EnzyRCD (e.g.: this constrained docking) should take and output /structure objects or anything **insensitative to the change of engine**. (or in extreme cases that you think this function will never have any need to use another engine)

shaoqx: And for parts that only EnzyRCD uses, you can have them stored in a submodule of `/preparation`

enzy_htp/__init__.py
```diff
@@ -48,4 +47,11 @@
 end users and instead exists for developers to access external software. See enzy_htp/_interface/interface.py
 for full class defintion."""
 
-from .analysis import ( electric_field )
+config = interface.config()
```
shaoqx, line 50: wont this overwrite what's in line 38.

shaoqx:
This is such an exciting PR! Love all the new interfaces and EnzyRCD!

Please check the comments above. I'm mainly not very comfortable with how files and functions under `/preparation` are set up. But I understand this might not be the final version.

shaoqx:
Also please address all the conflicts as a final commit after commits addressing the comments. Thank you! This will help track the code change.

enzy_htp/preparation/reactive_docking.py
```diff
@@ -0,0 +1,1285 @@
...
+from enzy_htp.structure import PDBParser, Structure
+from enzy_htp.core import file_system as fs
+
+#TODO(CJ): I should go through and use existing values if cache mode is selected
+#TODO(CJ): need to be able to specify the size of the grid
+
+
+def dock_reactants(structure: Structure,
+                   reactants: List[str],
+                   constraints: List[RosettaCst] = None,
```
shaoqx, line 30: Is it possible to decouple this input from Rosetta? e.g.: create a general constrain class that can be convert to the format of different software inside their own interfaces

enzy_htp/preparation/reactive_docking.py
```diff
@@ -0,0 +1,1285 @@
...
+                   mutations: List[List[mm.Mutation]] = None,
```
shaoqx, line 31: I still dont complete get the idea of how have mutation here is different from doing mutation first and then run EnzyRCD again.

shaoqx: Dont need to address it here. We can chat about it later.

shaoqx: It should be the same if you move the mutation component here and put mutate_stru() in the main script (of apps enabled by this module) after calling dock_reactants().

shaoqx: It's also tricky to determine whether we want to mutate first or dock first. I think mutate first is a better idea as it is how nature works and there are extreme cases that a mutation would completely disable the old binding site.

enzy_htp/_config/__init__.py
chrisjurich: Agreed; I would re-write it myself so that it is safer but it is somewhat low priority for now. I will update in the future when I have time.

chrisjurich: I went ahead and changed this so that it is handled by a function in `enzy_htp._config.Config.load_config()` class. Solves the security problem with `exec()` and is generally cleaner.

enzy_htp/_interface/pymol_interface.py
chrisjurich: good to know! sorry for duplciation.

enzy_htp/preparation/constrained_docking.py
chrisjurich: yes I will make sure everything is in `Structure()`/`Ligand()` objects. For now I just wanted to get it working.

enzy_htp/preparation/reactive_docking.py
chrisjurich: sure!

enzy_htp/preparation/reactive_docking.py
chrisjurich: So this is here in the event that a constraint and mutation overlap. If that is the case then the system is re-run separately with a set of modified inputs.

chrisjurich:
Made some changes; namely removed the call to `exec()`. I think it is best to merge the changes now and then resolve some of the design elements as the Ligand class is improved.

enzy_htp/preparation/align_ligand.py
chrisjurich: Due to how this alignment algorithm functions, the input file format HAS to be `.mol2`. This is because the function requires connectivity, bond type, and PDB-style atom naming. There is no other format that allows for this. I am not sure where this should go. The issue is that this function is not a true "alignment"; it uses PDB atom naming to align heavy atoms. I probably just need to put a more descriptive name.

enzy_htp/preparation/align_ligand.py
chrisjurich: I agree. All access to `rdkit` should be done through the `RDKItInterface()`

enzy_htp/preparation/cluster.py
chrisjurich: The one in `PyMolInterface()` is the correct one to use. Let me remove this.

enzy_htp/preparation/reactive_docking.py
chrisjurich: I agree. For now I was just doing everything in files but I will change the output format eventually.

enzy_htp/__init__.py
chrisjurich: It will! It's a no-op but still dumb. I have removed it.

enzy_htp/preparation/align_ligand.py
shaoqx: I think `.pdb` also fits these requirements lol.
For what you describe, you can put this in `structure_operation/ligand.py` or `preparation/reactive_docking` (is it only going to ever be used by EnzyRCD)

enzy_htp/preparation/constrained_docking.py
shaoqx: Cool! However, the idea of `/structure` is to only have them store/provide_access_to the data. Here are some idea about where functions should go:
- Addressing a high-level scientific problem (like docking or protonation) -> "the 4 high-level modules"
- Operation of the structure that is going to be repeatedly used to address different scientific problems -> `structure.structure_operation`
- Operation of the structure that is only going to be used by EnzyRCD -> `preparation.reactive_docking.xxx`

enzy_htp/preparation/reactive_docking.py
shaoqx: Cool. In this case, we can just leave the mutant input only used for that purpose.

shaoqx (approved):
Great changes! Very exciting!
Please file an issue link to this PR and address the remaining design comments in the next PR.

PR #140, opened 2023-12-06
Out-of-the-box installation with Conda and Python 3.9.

SwordJack:
Dear colleagues,

This commit is to enable out-of-the-box installation of EnzyHTP (refactor) to deal with the conflicts among packages and painful `Solving Environments` loading. An `environment.yml` file is provided with specified dependency versions and builds, as well as a `conda-install` script in the `dev-tools` directory.

Feel free to let me know if you have any questions.

Best wishes!


environment.yml
```diff
@@ -0,0 +1,266 @@
+name: enzyhtp
+channels:
+  - salilab
```
shaoqx, line 3: what is this for?

dev-tools/conda-install
```diff
@@ -0,0 +1,87 @@
...
+done
+
+# Get the conda version
+CONDA_VERSION=$(conda --version | cut -d' ' -f2)
+echo "Conda Version: $CONDA_VERSION"
+
+# Remove the build directory if it exists
+if [[ -d build ]]; then
+    rm -r build;
+fi
```
shaoqx, line 52: Is this necessary?

dev-tools/conda-install
```diff
@@ -0,0 +1,87 @@
...
+
+# Function to update the conda environment
+update_env() {
+    echo "Updating conda environment: $target_env_name"
+    conda env update -n "$target_env_name" --file environment.yml
+    conda develop .
```
shaoqx, line 58: We will eventually use `conda-build` to do this fyi

shaoqx: Also when this func is called in the `else` block, can `conda develop .` add EnzyHTP to the right env?

shaoqx:
Great job! Looks good for a temp use before finishing the refactoring. Please address these comment and I will approve the PR.

shaoqx (changes requested):

environment.yml
SwordJack: This is the channel to install `modeller` package. See https://anaconda.org/salilab/modeller

dev-tools/conda-install
SwordJack: It's assumed that this block copied from a template elsewhere is to clear up existing build directory so as to address some potential conflicts. I'm not sure when it would be triggered, but at least it will not cause error under any circumstances. Thus I kept it.

dev-tools/conda-install
SwordJack: > We will eventually use `conda-build` to do this fyi

Of course! Let me know when you are to use `conda-build`.

shaoqx: I think `conda develop .` will install `.` in the current env. Correct me if Im wrong

shaoqx: Oh I saw you addressed it.

shaoqx (approved):
Looks good! Ready for merge

PR #147, opened 2023-12-23
Chrisjurich/reactive docking pr2

chrisjurich:
This is a major update to the reactive docking API as well as internals in the `structure.structure_constraint` and `structure.structure_region` sub-modules. Brief descriptions are below:

+ `structure.structure_constraint` allows for the definition of generic `StructureConstraint` classes that define a variety of geometric constraints between residues and atoms. Basic infrastructure for constraint I/O was implemented

+ `structure.structure_region` allows for the creation of subregions in the Pose. It is mostly used for QM clustering. While variation of this have existed before, they relied on pymol and were prone to failure. The new version uses internal functions and does not require pymol or the corresponding interface.


enzy_htp/preparation/__init__.py
```diff
@@ -11,6 +11,6 @@
 from .conformer_generation import generate_conformers
 from .protonate import protonate_stru
 from .clean import remove_solvent
-from .reactive_docking import dock_reactants
+from .reactive_docking import dock_reactants, _system_charge, _qm_minimization
```
shaoqx, line 14: This looks pretty weird. Is _system_charge and _qm_minimization an API that user should use or an internal method that only used by .reactive_docking (as "_" implies)

enzy_htp/structure/ligand.py
```diff
@@ -100,6 +100,22 @@ def add_conformer(self, points:List[Tuple[float,float,float]] ) -> int:
 
         return self.n_conformers()
 
+    def get_ligand_conformer(self, idx:int) -> Ligand:
```
shaoqx, line 103: This will be merged into the a method of StructureEnsemble in the future FYI.
Probably add a TODO about this next to this func?

enzy_htp/structure/chain.py
```diff
@@ -1,7 +1,6 @@
 """Definition for the Chain class. Chains primarily store Residue() objects and organize them
 within the overall structure of an enzyme.
-Author: Qianzhen (QZ) Shao <shaoqz@icloud.com>
-Author: Chris Jurich <chris.jurich@vanderbilt.edu>
+Author: Qianzhen (QZ) Shao <shaoqz@icloud.com> Author: Chris Jurich <chris.jurich@vanderbilt.edu>
```
shaoqx, line 3: Why is this formatting lol?

enzy_htp/structure/structure.py
```diff
@@ -614,6 +614,48 @@ def assign_ncaa_chargespin(self, net_charge_mapper: Dict[str, Tuple[int, int]]):
 
     #endregion
 
+    def get_residue(self, key:str) -> Residue:
```
shaoqx, line 617: A function that does the same is `find_residue_with_key()`. Check it out. It takes a tuple as arg tho. We can either make a if flow inside of it or a dispatch.

enzy_htp/structure/structure.py
```diff
@@ -614,6 +614,48 @@ def assign_ncaa_chargespin(self, net_charge_mapper: Dict[str, Tuple[int, int]]):
...
+        for res in self.residues:
+            if res.parent.name == chain and res.idx == rnum:
+                return res
+        else:
+            _LOGGER.error(f"Unable to locate residue {key} in Structure. Exiting...")
+            exit( 1 )
+
+
+
+    def get_atom(self, key:str) -> Atom:
```
shaoqx, line 643: I see what you want here.

I would like to design this as `Structure().get(key: str) -> Union[Chain, Residue, Atom]`

enzy_htp/structure/atom.py
```diff
@@ -128,7 +128,16 @@ def element(self):
                 return self.parent.element
             else:
                 # case: in ligand atoms are named like this H1
-                return re.match("^[A-Z][a-z]?", self.name).group()
+                for ch in self.name:
+                    if not ch.isnumeric():
+                        return ch
```
shaoqx, line 133: Oh I see there are names starting with number. But the current logic also dont work for atoms like Mg or Cl, which will be M and C returned

enzy_htp/structure/structure_region/api.py
```diff
@@ -0,0 +1,31 @@
+
```
shaoqx, line 1: This is correct for this PR? It seems from the desc of the PR StructureRegion should be kinda implemtated. 

I'm ok with just a framework but just to make sure.

enzy_htp/structure/structure_constraint/api.py
```diff
@@ -1,43 +1,477 @@
...
+StructureConstraints are meant to define flexible, non-package specific relationships that can be translated 
+in between different software packages. Factory functions for constraints are also included. The contained 
+classes are:
+
+    + StructureConstraint: The abstract base class.
+    + CartesianFreeze: Specific atoms will be frozen.
+    + DistanceConstraint: Constraint that specifies a distance.
+    + AngleConstraint: Constraint that specifies an Angle.
+    + DihedralConstraint: Constraint that specifies a dihedral.
+    + ResiduePairConstraint: Specifies a constraint set between two residues in a style similar to Rosetta EnzDes.
```
shaoqx, line 11: Making them seperate descendant classes is absolutly the right move! In most cases, methods will be specialized for each specific types of constraints. Great job!

enzy_htp/structure/structure_constraint/api.py
```diff
@@ -1,43 +1,477 @@
...
 
 Author: Qianzhen (QZ) Shao, <shaoqz@icloud.com>
+Author: Chris Jurich <chris.jurich@vanderbilt.edu>
 Date: 2022-10-28
 """
+from __future__ import annotations
```
shaoqx, line 17: we probably dont want this as the python version is limited to >3.9 and it doesn't exist in other modules.

enzy_htp/preparation/__init__.py
chrisjurich: lol I definitely left this in here by accident. Let me remove it.

enzy_htp/structure/structure_constraint/api.py
```diff
@@ -1,43 +1,553 @@
...
+                target_value:float,
+                params:Dict[str, Any]
+                ):
+        """Ctor that exits if the supplied number of atoms is NOT correct."""
+        
+        self.atoms_ = atoms 
+        self.target_value_ = target_value 
+        self.params_ = deepcopy( params ) 
+
+        self['calc_method'] = params.get('calc_method', 'rosetta')
```
shaoqx, line 52: why does this exist? The StructureConstraint class itself should not know anything about the interface. It should know only know constraint as a math model.

enzy_htp/structure/structure_constraint/api.py
```diff
@@ -1,43 +1,553 @@
...
+
+        if not self.correct_num_atoms():
+            _LOGGER.error("Incorrect number of atoms supplied! Exiting...")
+            exit( 1 )
+
+
+    #TODO(CJ): add function that checks if topology and constraints are compatible
+    #TODO(CJ): will need to make a version of this that actually works for the ResiduePairConstraint
```
shaoqx, line 60: probably a typo here? Not it is not finished yet.

enzy_htp/structure/structure_constraint/api.py
shaoqx: (for decoupling purpose)

Anything related to rosetta should be in the RosettaInterface

enzy_htp/structure/chain.py
chrisjurich: Quite odd. Probably caused by a defective vim shortcut or something.

enzy_htp/structure/structure_constraint/api.py
```diff
@@ -1,43 +1,553 @@
...
+
+    def is_angle_constraint(self) -> bool:
+        """Is this an angle constraint?"""
+        return False
+
+    def is_dihedral_constraint(self) -> bool:
+        """Is this a dihedral constraint?"""
+        return False
+
+    def is_residue_pair_constraint(self) -> bool:
```
shaoqx, line 87: This method should be in RosettaInterface that takes StructureConstraint() as an input arg

shaoqx: Because this logic is based on the knowledge of Rosetta

enzy_htp/structure/ligand.py
chrisjurich: Fixed this and added a TODO note about it.

enzy_htp/structure/structure_constraint/api.py
shaoqx: This is comparable to putting a `escort` method inside of a `cargo` class instead of `ship`, `truck`, `airplane` classes.

enzy_htp/structure/atom.py
chrisjurich: Good point. Replaced the logic and updated it in a commit to not be susceptible to the bug.

enzy_htp/structure/structure_region/api.py
chrisjurich: It is correct. Will reorganize at a future point.

enzy_htp/structure/structure_constraint/api.py
chrisjurich: Thanks!

enzy_htp/structure/structure_constraint/api.py
```diff
@@ -1,43 +1,553 @@
...
+
+    def change_topology(self, new_topology:Structure) -> None:
+        """Switch the associated topology and atoms for a given Constraint. Maps the atom keys
+        as described in Atom.key() and Structure.get_atom().
+
+        Args:
+            new_topology: new structure to translate atoms to.
+
+        Returns:
+            Nothing.
```
shaoqx, line 121: This seems does nothing lol

enzy_htp/structure/structure_constraint/api.py
chrisjurich: It is just a method of calculating the constraint. It can be renamed to something else. Since constraints are most often implemented in MD/QM packages, I figured it would be best to use potentials/functions that are accurate to those packages.

enzy_htp/structure/structure_constraint/api.py
chrisjurich: yes. not finished yet

enzy_htp/structure/structure_constraint/api.py
chrisjurich: Fixed and removed!

enzy_htp/structure/structure_constraint/api.py
chrisjurich: I think it is fine to be here. `ResiduePairConstraint` is a type of `StructureConstraint` and is just a collection of other types of constraints.

enzy_htp/structure/structure_constraint/api.py
shaoqx: I argee. I found this is necessary if params are assigned.

enzy_htp/structure/structure_constraint/api.py
chrisjurich: huh? wym

enzy_htp/structure/structure_constraint/api.py
shaoqx: But here is a problem.

In the case I have MD -> QM(xtb) -> QM(Gaussian) in my workflow and I want to apply the same set of constraints, I can no longer specific the params inside of the StructureConstraint object.

shaoqx: Could this be a good idea?

We allow each XXXInterface to have their on descendant StructureConstraint class like RosettaConstraint that only have the additional .params attribute. If user want to specify params for each software in the workflow, they can construct these software specific children classes and give them to the interface.


The function in the interface that apply these constraint will have a default set of params that is used when params in the StructureConstraint() dont exist. (In QM and MD this is the most common cases that only a default value/pattern is always used.)

enzy_htp/structure/structure_constraint/api.py
shaoqx: cool

enzy_htp/structure/structure_constraint/api.py
shaoqx: It does not raise an error but just returns a boolean

enzy_htp/structure/structure_constraint/api.py
```diff
@@ -1,43 +1,552 @@
...
+        Returns:
+            Value corresponding to the specified key.
+        """
+
+        if key not in self.params_:
+            _LOGGER.error(f"The attribute '{key}' is not in the constrained geometry params. Exiting...")
+            exit( 1 )
+        
+        return self.params_[key]        
+
```
shaoqx, line 184: Should be True right?

enzy_htp/structure/structure_constraint/api.py
```diff
@@ -1,43 +1,552 @@
...
+        return True
+
+    def correct_num_atoms(self) -> bool:
+        """A distance constraint should have two atoms."""
+        return len(self.atoms) == 2
+
+    def current_geometry(self) -> float:
+        """The cartesian distance between two 3D points."""
+        return self.atoms[0].distance_to(self.atoms[1])
+
```
shaoqx, line 236: we can add this under Atom

enzy_htp/structure/structure_constraint/api.py
```diff
@@ -1,43 +1,552 @@
...
+    def correct_num_atoms(self) -> bool:
+        """A dihedral constraint should have 4 atoms."""
+        return len(self.atoms) == 4
+
+    def current_geometry(self) -> float:
+        """Measurement for the current dihedral between the 4 atoms in the constraint."""
+        p0 = np.array(self.atoms[0])
+        p1 = np.array(self.atoms[1])
+        p2 = np.array(self.atoms[2])
+        p3 = np.array(self.atoms[3])
```
shaoqx, line 277: We can add it to Atom too.

Also we should use `core.math_helper.get_dihedral()` for this

enzy_htp/structure/structure_constraint/api.py
shaoqx: Oh I dont mean `ResiduePairConstraint` but the `score_energy` method should be inside of `RosettaInterface`

enzy_htp/structure/ligand.py
shaoqx: Nice

enzy_htp/structure/atom.py
shaoqx: Nice change!

But the current logic still dont work with atom names like CH3 in e.g. FAH ligand.

shaoqx: I also found in a ligand called DCE, Cl's name is CL1, CL2 which is kind hard to deduce using just the name. The ultimate solution would be using a mapper that record all these.

Do you mind if I push some unit tests and changes related to this on this branch?

enzy_htp/structure/structure.py
shaoqx: We can either leave a TODO here or I can help apply these changes if it is too much to change

enzy_htp/structure/structure_region/api.py
shaoqx: Cool!

enzy_htp/structure/structure.py
```diff
@@ -597,6 +582,16 @@ def add(self, target: List[Residue], # pylint: disable=function-redefined
             for ch in self:
                 ch.sort_residues()
 
+    @dispatch
+    def absolute_index(self, target:Residue, indexed:int=0) -> int:
```
shaoqx, line 586: What is this dispatching against?

enzy_htp/structure/atom.py
```diff
@@ -276,6 +281,26 @@ def is_connected(self) -> bool:
         """check if self is in the connected state"""
         return self._connect is not None
 
+    @property
```
shaoqx, line 284: Probly move it to the region with all the other @property? I can help on this.

enzy_htp/structure/ligand.py
```diff
@@ -46,6 +46,7 @@ def __init__(self, residue_idx: int, residue_name: str, atoms: List[Atom], paren
 
         self.bonds = kwargs.get('bonds', list())  # TODO change this to a more general represetation
         self.conformer_coords = list() # TODO change this to StructureEnsemble (we can leave a reference_var here tho)
+        self.placement_method_ = ''
```
shaoqx, line 49: May be make it None for the convention?

The idea for Structure and Ligand is to only have the math model recorded. But I see there is not really a better place to record the source of placement. Would a self.mimo attribute a good idea for these? So that we can distinguish those that should not assumed to exist but just used in some protocols. (similar to what I did in job_manager.ClusterJob)

shaoqx: Just add a TODO here is fine for me.

enzy_htp/structure/residue.py
```diff
@@ -293,6 +306,27 @@ def shift(self, point:List[float] ) -> None:
             aa.coord = updated
 
     #endregion
+    #TODO(CJ): the documentation for this part
+    def clash_count(self, other:Residue, radius:float=2.0, ignore_H:bool=True) -> int:
```
shaoqx, line 310: put this to Getter-Prop region?

enzy_htp/structure/atom.py
shaoqx: Also probably for consistency purpose, we can rename it as key_str as Residue did.

shaoqx (changes requested):
Great changes! Some minor formatting and design issue left. I can help on "region" related stuff and element unit tests if you don't mind :)

shaoqx:
I pushed several commits to the branch for fixing some bug and unit tests.
Sorry for doing it before having your consent, i can revert it if you dont like them.

enzy_htp/structure/atom.py
shaoqx: I fixed it and filed a TODO for the CL case. Also added some unit tests

enzy_htp/structure/structure_constraint/api.py
shaoqx: Wait I see lol. I dont know why the location shifted. It must be after new commits the lines are different. It was for score_energy()

enzy_htp/structure/structure_constraint/api.py
shaoqx: Oh the location of this is also wrong. I will make a new comment.

enzy_htp/structure/structure_constraint/api.py
```diff
@@ -1,43 +1,552 @@
...
+        self.torsionAB_ = torsionAB
+
+        atoms = set()
+        for (cst_name,cst) in self.child_constraints:
+            if cst is None:
+                continue
+            atoms.update( cst.atoms)
+        
+        self.atoms_ = list(atoms)
+        self.correct_num_atoms()
```
shaoqx, line 339: This seems does nothing lol

enzy_htp/structure/structure_constraint/api.py
```diff
@@ -1,43 +1,552 @@
...
+                cst.change_topology(new_topology)
+
+        atoms = set()
+        for (cst_name,cst) in self.child_constraints:
+            if cst is None:
+                continue
+            atoms.update( cst.atoms)
+        
+        self.atoms_ = list(atoms)
+        self.correct_num_atoms()
```
shaoqx, line 379: Same here

enzy_htp/structure/structure_constraint/api.py
```diff
@@ -1,43 +1,552 @@
...
+
+class CartesianFreeze(StructureConstraint):
+    """Specialization of StructureConstraint() for Atoms() that are frozen in Cartesian space. Many
+    of the methods yield junk values since no measurement is needed for this Class.
+    """
+    
+    def __init__(self, atoms:List[Atom] ):
+        super().__init__(self, atoms, 0.0, dict())
+     
+    def is_cartesian_freeze(self) -> bool:
```
shaoqx, line 206: Should be True right?

enzy_htp/structure/structure_constraint/api.py
```diff
@@ -1,43 +1,552 @@
...
+        a = np.array(self.atoms[0].coord)
+        b = np.array(self.atoms[1].coord)
+        c = np.array(self.atoms[2].coord)
+        
+        ba = a - b
+        bc = c - b
+        
+        cosine_angle = np.dot(ba, bc) / (np.linalg.norm(ba) * np.linalg.norm(bc))
+        return np.degrees(np.arccos(cosine_angle))
+
```
shaoqx, line 259: We can add this under Atom too

enzy_htp/structure/structure_constraint/api.py
shaoqx: This is for current_geometry()

enzy_htp/structure/structure_constraint/api.py
shaoqx: I marked TODO but still want to see how do you see this

enzy_htp/structure/structure_constraint/api.py
shaoqx: I marked TODO but still want to see how do you comment on this

enzy_htp/structure/structure.py
shaoqx: I changed it in a commit

enzy_htp/structure/structure.py
shaoqx: I merged it to get()

shaoqx (approved):
Great PR! Amazing changes in the Rdock API, StructureConstraint, and StructureRegion.
There are still several open discussions left but none of them are emergency.

shaoqx:
Merging. Making unresolved discussions issues,

enzy_htp/structure/structure_constraint/api.py
shaoqx: I thought about it again, 3 potential solution:
            - add a self.mimo for non-standard data storage (same as current solution)
            - make sth like AmberCartesianFreeze decsendant classes for each class and for
              each interface. Where user can further define parameters.
              e.g.: cons: AmberCartesianFreeze = in_amber(CartesianFreeze(), weight=2.0) 
            - add a mapper that contain parameters for EVERY software of support. user can just edit
              This mapper to change the parameter.

shaoqx: I kinda lean to the 2rd solution.
The 1st one adds too much burden to the developer and user.

shaoqx: I want to do the 2nd solution. Applying for your consent! ✋ 
In this case, self.params will be like this
{
"rosetta" : {"xxx":1.0},
"amber" : {"xxx":1.0},
...
}
and the value will be different for each constraint type. i.e. the supported software and weight keywords/value will be different

PR #155, opened 2024-01-12
Fixing a number of deletion issues and name chainges in various `__init__.py` files

chrisjurich:
Clerical fixing of various import names and deleting things here and there.


shaoqx (approved):
Good catch! Approved

PR #157, opened 2024-01-15
Chrisjurich/fix eletronic spelling

chrisjurich:
Turns out that `ElectronicStructure` was mispelled as `EletronicStructure` throughout a few files. This commit and PR just changes the spelling to be correct.


shaoqx (approved):
lol nice catch

PR #162, opened 2024-01-30
Chrisjurich/structure region fix

chrisjurich:
This PR contains a variety of items but is focused on getting the `StructureRegion` and specifically the residue capping functionality up and running. Also some changes to the reactive docking protocol but they are not very relevant to much.


enzy_htp/_interface/amber_interface.py
```diff
@@ -683,14 +682,12 @@ def run(self, input_data: Union[MolDynParameter, MolDynResult]) -> MolDynResult:
...
 
-        result = MolDynResult(
-            traj_file = traj_path,
-            traj_parser = traj_parser, 
-            traj_log_file = mdout_path,
-            traj_log_parser = traj_log_parser, 
-            last_frame_file = mdrst_path,
-            last_frame_parser = last_frame_parser,
-            source="amber",
+        result = AmberMDResultEgg(
```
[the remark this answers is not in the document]
shaoqx: Please change this back

enzy_htp/_interface/amber_interface.py
```diff
@@ -755,7 +752,8 @@ def translate(self, result_egg: AmberMDResultEgg) -> MolDynResult:
         # clean up
         clean_up_target = ["./mdinfo"]
         if not self.keep_in_file:
-            clean_up_target.extend(result_egg.parent_job.mimo["temp_mdin"])
+            if result_egg.parent_job is not None:
+                clean_up_target.extend(result_egg.parent_job.mimo["temp_mdin"])
```
shaoqx, line 756: result egg have to have a parent job

enzy_htp/_interface/amber_interface.py
```diff
@@ -1485,16 +1519,28 @@ def _parse_md_config_dict_to_raw(self, md_config_dict: Dict) -> Dict:
...
             disang_content_list = []
             for cons in geom_cons:
-                raw_rs_dict = self._parse_cons_to_raw_rs_dict(cons)
-                disang_content_list.append(raw_rs_dict)
+                if cons.is_residue_pair_constraint():
+                    for (_, cst) in cons.child_constraints:
+                        disang_content_list.append( self._parse_cons_to_raw_rs_dict(cst) )
+                else:                    
+                    raw_rs_dict = self._parse_cons_to_raw_rs_dict(cons)
+                    disang_content_list.append(raw_rs_dict)
```
shaoqx, line 1543: I guess a good treatment would be if residue pair constraint is found, extend the geom_cons list by .child_constraints

shaoqx: The current way is pretty hard to read and repeatative

enzy_htp/_interface/amber_interface.py
```diff
@@ -1251,6 +1249,42 @@ def get_amber_atom_index(self, atoms: List[Atom]) -> List[int]:
         aid_mapper = self.get_amber_index_mapper(stru)
         return [aid_mapper["atom"][at] for at in atoms]
 
+
+    def rename_atoms(self, stru: Structure) -> None:
```
shaoqx, line 1253: The design for atom names is that:
- we use a naming conventions as a standard in EnzyHTP
- file parsers needs to parse non-standard names to standard ones when reading and vice versa

shaoqx: In this case, we should not change atom names in Structure() (it is also dangerous as some functions relies on it to figure out the topology.). This function should be in e.g.: pdb_io and we can add options for the PDB style of input or output.

enzy_htp/_config/armer_config.py
```diff
@@ -71,6 +71,22 @@ def required_py_modules(self) -> List[str]:
...
+            return getattr(self, key)
+
+    def __setitem__(self, key: str, value: Any) -> None:
+        """Setter that enables [] accession of ARMerConfig() attributes with value validation."""
+        if key.count("."):
+            key1, key2 = key.split(".")
+            ARMerConfig.__dict__[key1][key2] = value
+        else:
+            setattr(self, key, value)
+
```
shaoqx, line 89: These are in BaseConfig right?

enzy_htp/structure/structure_constraint/create_constraint.py
```diff
@@ -0,0 +1,310 @@
...
+            _LOGGER.error(f"Inconsistent topology in the target list. {cons.topology} -vs- {ref_cons.topology} ")
+            raise ValueError
+    # merge atoms
+        result.atoms.extend(cons.atoms)
+    result.atoms = list(set(result.atoms))
+    return result
+# endregion
+
+
+def freeze_hydrogen_bonds(sr:StructureRegion, params:Dict=None ) -> List[DistanceConstraint]:
```
shaoqx, line 298: We can rename it to create_hydrogen_bond_freeze to unify the API

enzy_htp/structure/structure_operation/general.py
```diff
@@ -126,4 +128,19 @@ def align_atom_order_in_each_residue(stru: Structure, ref_stru: Structure) -> No
             atom_indexes = res.atom_idx_list
             atom_indexes.sort()
             res.renumber_atoms(atom_indexes)
-    return stru
\ No newline at end of file
+    return stru
+
+
+def update_coordinates(stru:Structure, ref_stru:Structure, allow_index_mismatch:bool=True) -> None:
```
shaoqx, line 134: We have apply_geom already in Structure. It also check the topology to avoid potential hard to find bugs.

enzy_htp/structure/structure_region/api.py
```diff
@@ -61,10 +63,25 @@ def atoms(self) -> List[Atom]:
...
 
     @atoms.setter
-    def atoms(self, val: List[Atom]):
+    def atoms(self, val: List[Atom]) -> None:
         """setter for atoms_"""
         self.atoms_ = val
 
+        for atom in self.atoms:
+            if atom.parent.parent is None:
+                atom.parent.parent = self
```
shaoqx, line 72: We should not set the parent here. In this case, it will lose track to the Structure() context (i.e.: the self.topology will no longer work)

shaoqx: We need this context in many places especially when dealing with ElectronicStrutcure.

enzy_htp/structure/structure_region/api.py
```diff
@@ -61,10 +63,25 @@ def atoms(self) -> List[Atom]:
...
+        #TODO(CJ): set the parent.
+
+    def add_atoms(self, new_atoms:List[Atom]) -> None:
+        #TODO(CJ): documentation!
+
+        self.atoms.extend(new_atoms)
+
+        for atom in self.atoms:
+            if atom.parent.parent is None:
+                atom.parent.set_parent(self)
```
shaoqx, line 82: same as above

enzy_htp/structure/structure_region/api.py
```diff
@@ -174,13 +192,19 @@ class is only topology related. If create an obj for
...
             res = atom.parent
-            if not isinstance(res, ResidueCap):
-                result.add(res)
-        return list(result)
+            if isinstance(res, ResidueCap):
+                continue
+
+            if res not in unique:
+                unique.add(res)
+                result.append(res)
```
shaoqx, line 205: `.add()` and `not in` is duplicated in terms of making element unique.

shaoqx: Or you are making the order not the same as how atom appears?

enzy_htp/structure/structure_region/api.py
```diff
@@ -224,10 +248,34 @@ def involved_residues_with_free_terminal(self) -> Dict[str, List[Residue]]:
                     result["c_ter"].append(res)
                 if n_side_res is not None and n_side_res not in region_residues:
                     result["n_ter"].append(res)
+            elif res.is_ligand():
+                continue
```
shaoqx, line 252: Is this for potential treatment when truncating a ligand?

enzy_htp/structure/structure_region/api.py
```diff
@@ -224,10 +248,34 @@ def involved_residues_with_free_terminal(self) -> Dict[str, List[Residue]]:
...
+
+    def needs_cterm_cap(self, res: Residue) -> bool:
+
+        if res.is_ligand():
+            return False
+
+        region_residues = self.involved_residues
+        c_side_res = res.c_side_residue()
+
+        return c_side_res is not None and c_side_res not in region_residues
```
shaoqx, line 276: The involved_residues_with_free_terminal will replaces these I think

enzy_htp/structure/structure_region/api.py
```diff
@@ -303,6 +351,10 @@ def topology(self) -> Structure:
                 return atom.root()
     # endregion
 
+
+    def root(self, n):
+        return self.topology
```
shaoqx, line 356: Remove this I guess?

enzy_htp/structure/structure.py
```diff
@@ -771,13 +778,22 @@ def add(self, target: List[Residue], # pylint: disable=function-redefined
 
     @dispatch
     def absolute_index(self, target:Residue, indexed:int=0) -> int:
```
shaoqx, line 780: What is this dispatching against?

enzy_htp/structure/structure_region/capping.py
```diff
@@ -265,120 +574,25 @@ def cap_residue_free_terminal(
...
+    "H" : HCap,
+    "NHCH3": NHCH3Cap,
+    "COCH3": COCH3Cap,
+    # common names for redundancy
+    "methyl": CH3Cap,
+    "hydrogen": HCap,
+    "methylamide": NHCH3Cap,
+    "acetate": COCH3Cap
 }
+"""Mapper that converts a cap name into a constructor. Both the chemical formula and common names for each method are provided."""
```
shaoqx, line 598: Nice!

enzy_htp/structure/structure_region/capping.py
```diff
@@ -50,124 +62,222 @@ def capping_with_residue_terminals(raw_region,
...
     if return_copy:
         new_region = raw_region.clone()
-        new_region.atoms.extend(add_atoms)
+        new_region.atoms = new_atoms 
         return new_region
     else:
-        raw_region.atoms.extend(add_atoms)
-
-# endregion
+        raw_region.atoms = new_atoms
```
shaoqx, line 81: I guess they works the same but the API from StructureRegion is different. Do you mind if we change it back?

enzy_htp/structure/structure_region/capping.py
```diff
@@ -50,124 +62,222 @@ def capping_with_residue_terminals(raw_region,
...
-        socket_atom:
-            the socket from original stru to pluging the cap.
-            the cap is replacing the atom.
-        plug_atom:
-            the pluging in atom in the cap
+    do not exist in the real structure.
+
+    Attributes:
+        link_residue: The Residue() being capped.
+        link_atom: THe terminal Atom() that needs capping specifically.
```
shaoqx, line 94: 911 I report a typo here

enzy_htp/structure/structure_region/capping.py
```diff
@@ -50,124 +62,222 @@ def capping_with_residue_terminals(raw_region,
...
             _LOGGER.error("link atom must be one of the link residue atoms")
             raise ValueError
-        self.atoms = atoms
-        self.plug_atom = plug_atom
-        if plug_atom not in self.atoms:
-            _LOGGER.error("plug atom must be one of the atoms")
-            raise ValueError
 
-        self._idx = None
-        self._name = None
```
shaoqx, line 111: We need to keep this to make logging normal if not working methods from Residues is called

enzy_htp/structure/structure_region/capping.py
```diff
@@ -50,124 +62,222 @@ def capping_with_residue_terminals(raw_region,
...
         self.set_ghost_parent()
-        self.set_children(atoms)
-    
+
+        self.anchor()
+
+
+    @abstractmethod
+    def anchor(self) -> None:
+        """Function that moves the ResidueCap()'s Atom()'s to be in the correct position for the type of Cap that it is."""
```
shaoqx, line 132: This desc is a bit vague I think. especially the "for the type of Cap that it is." part

enzy_htp/structure/structure_region/capping.py
```diff
@@ -50,124 +62,222 @@ def capping_with_residue_terminals(raw_region,
...
+        """Gets the index of the link_residue."""
+        return self.link_residue.idx
 
     @property
-    def cap_bond_vector(self) -> np.array:
-        """return a vector that defines the direction of the capping bond"""
-        result = np.array(self.socket_atom.coord) - np.array(self.link_atom.coord)
-        result /= np.linalg.norm(result)
-        return result
+    def chain(self) -> Chain:
```
shaoqx, line 176: Chain is not imported

enzy_htp/structure/structure_region/capping.py
```diff
@@ -50,124 +62,222 @@ def capping_with_residue_terminals(raw_region,
...
+
+
+class HCap(ResidueCap):
+    """A ResidueCap() which uses a -H (hydrogen) as a cap. Adds a total of 1 atom1, with the 
     
-    def restore_default_coord(self):
-        for atom in self.atoms:
-            atom.coord = self.DEFAULT_COORDS[atom.name]
-
-    def apply_to_geom(self, geom: Structure) -> ResidueCap:
```
shaoqx, line 143: StructureRegion use this function for atoms_from_geom and thus for single_point to deal with ensemble input. Please restore it.

test/structure/structure_region/test_capping.py
```diff
@@ -24,35 +25,184 @@ def test_capping_with_residue_terminals(helpers):
...
-        str(len(atoms)),
-        "",]
-    for at in atoms:
-        lines.append(f"{at.element} {at.coord[0]} {at.coord[1]} {at.coord[2]}")
-    fs.write_lines(test_file, lines)
-    assert helpers.equiv_files(test_file, answer_file, consider_order=False)
-
-    fs.safe_rm(test_file)
-
-def test_residue_cap_deepcopy():
```
shaoqx, line 46: please also keep this test

test/structure/structure_region/test_capping.py
```diff
@@ -24,35 +25,184 @@ def test_capping_with_residue_terminals(helpers):
...
+    sele = stru_sele.select_stru(
+        test_stru, "resi 3")
+
+    test_region = stru_regi.StructureRegion(atoms=sele.atoms)
+
+    caps = set()
+
     capping.capping_with_residue_terminals(test_region)
-    # test
-    atoms = test_region.atoms
```
shaoqx, line 33: Please also make this test work. I can help if you need!

enzy_htp/structure/structure_region/capping.py
shaoqx: Please make `test/structure/structure_region/test_api.py::test_atoms_from_geom()` work

enzy_htp/structure/structure_region/capping.py
```diff
@@ -179,72 +289,271 @@ class CH3Cap(ResidueCap):
...
+        cap_bond_distance = self.CAP_BOND_DISTANCE[self.link_atom.name]
 
-    CAP_BOND_DISTANCE = {
-        "N" : 1.1,
-        "C" : 1.0,
+        for aa in self.atoms:
+            pt = np.array(aa.coord)
+            pt = np.transpose(np.matmul(rot_mat, np.transpose( pt  )))
+            pt += (direction*cap_bond_distance) + p1 
+            aa.coord = pt
```
shaoqx, line 342: We should generalize it and put the math part in core/math_helper.py

shaoqx: The current anchor dont work when the cap is not in the initial position I think. i.e.: not from 1,0,0
But it is needed in apply_to_geom

shaoqx (changes requested):
Great PR! Many good changes to add! Please address these comments.
Also some unit tests fails after the PR. Please make them work, specifically, unit tests from
- amber_interface.py
- structure_constraint
- structure_region
- quantum

enzy_htp/_interface/xtb_interface.py
```diff
@@ -8,28 +8,241 @@
...
+            mo_parser = '',
+            source="xtb",
+        )
+
+    def translate(self, result_egg: XTBQMResultEgg) -> ElectronicStructure:
+        """TODO(CJ)"""
+
+        assert False
+
+class XTBOptimizationEngine(XTBSinglePointEngine, QMOptimizationEngine):
```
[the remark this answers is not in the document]
chrisjurich: I make it this way because this is also how the RosettaQM way works. I also just noticed that there isn't much difference besides the fact that this type of run will also use constraints and update coordinates

enzy_htp/_interface/amber_interface.py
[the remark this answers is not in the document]
chrisjurich: yeah idk what I was doing. I will just check out the `develop_refactor` version of the `amber_interface.py` file and have that fix all these problems. Sorry for the issue!

enzy_htp/_interface/amber_interface.py
chrisjurich: See above comment

enzy_htp/_interface/amber_interface.py
chrisjurich: Yeah I think so too. I just left it that way to get it to work for an example. I will fix it

enzy_htp/_interface/amber_interface.py
chrisjurich: This function is the result of some problems I have come across. Namely, the default naming scheme in EnzyHTP is essentially that of the Amber force field. As a result, things from Rosetta have to have special treatment before they are ready for MD. If there is a current way to do it, let me know. I guess I can try to use the original naming on the output from reactive docking but in general I don't think EnzyHTP is currently robust with respect to atom naming and conversion between packages.

enzy_htp/_interface/xtb_interface.py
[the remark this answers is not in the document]
chrisjurich: I think both could still inherit from `SimulationEngine`. I was just lazy and wrote a version o the `XTBInterface` functions that are the same for both run types and depend only on a flag for if they are geometry optimizations.

enzy_htp/core/clusters/accre.py
```diff
@@ -42,8 +42,8 @@ class Accre(ClusterInterface):
 
     # environment presets #
     AMBER_ENV = {
-        "CPU": "source /home/shaoq1/bin/amber_env/amber22.sh",
-        "GPU": "source /home/shaoq1/bin/amber_env/amber22.sh",
+        "CPU": "source ~/setup-accre.sh",
```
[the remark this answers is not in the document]
chrisjurich: I can change it back. This is kind of my argument for a config file for a cluster not being in the actual code distribution. I changed it to work for my specific setup. That being said I also changed it so that you can alter it from the script you are calling, but I still think this is a design aspect we should change.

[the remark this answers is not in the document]
chrisjurich: Yeah I will change this. This is part of my thought that the cluster config files should not be in the distribution of EnzyHTP. I guess this is somewhat fixed since I can change the `dict` directly

enzy_htp/quantum/api.py
```diff
@@ -222,14 +224,126 @@ def single_point(
...
+            stru_esm, qm_engine,
+        )
+    else:
+        _LOGGER.error(f"{parallel_method} is not a supported parallel_method")
+        raise ValueError        
+
+    return result
+
+
     raise Exception("TODO")
```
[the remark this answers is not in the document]
chrisjurich: Sure!

enzy_htp/structure/structure_region/capping.py
chrisjurich: good catch

test/structure/structure_region/test_capping.py
chrisjurich: Sure

test/structure/structure_region/test_capping.py
chrisjurich: Sure I can ig

enzy_htp/structure/structure_region/capping.py
chrisjurich: This is true that the cap must be in the default position. However, EnzyHTP does not have a robust mechanism for doing internal coordinates, so we have to resort to hacky implementations like this. To be clear, I do not think we should have a robust mechanism for internal coordinates, but I don't think there is a better way to deal with it.

enzy_htp/_interface/amber_interface.py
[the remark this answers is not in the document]
chrisjurich: restored with commit 1

enzy_htp/_interface/amber_interface.py
chrisjurich: restored with commit 1

enzy_htp/_interface/amber_interface.py
chrisjurich: reverted to previous with commit 1

enzy_htp/_interface/amber_interface.py
chrisjurich: What do you want me to do with this?

enzy_htp/chemical/residue.py
```diff
@@ -530,509 +530,599 @@
...
-            'HG13': 0.0791,
-            'CG2': -0.3192,
-            'HG21': 0.0791,
-            'HG22': 0.0791,
-            'HG23': 0.0791,
-            'C': 0.5973,
-            'O': -0.5679
-        },
-    },
+RESIDUE_CHARGE_MAP={
```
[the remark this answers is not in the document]
chrisjurich: I'll just revert this file for now. I don't think I actually need this for anything.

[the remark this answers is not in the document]
chrisjurich: File reverted with commit 2

enzy_htp/core/clusters/accre.py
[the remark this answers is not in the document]
chrisjurich: Reverted with commit 3

enzy_htp/quantum/api.py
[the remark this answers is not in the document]
chrisjurich: removed. Weird that I left his in lol

enzy_htp/quantum/api.py
```diff
@@ -222,14 +224,126 @@ def single_point(
...
+        constraints:List[StructureConstraint] = None,
+        capping_method: str = "res_ter_cap",
+        embedding_method: str= "mechanical", # TODO probably not a good default choice
+        parallel_method: str="cluster_job",
+        cluster_job_config: Dict= None,
+        job_check_period: int= 210, # s
+        job_array_size: int= 20,
+        work_dir: str="./QM_OPT",
+        keep_in_file: bool=False,
+        ) -> List[ElectronicStructure]:
```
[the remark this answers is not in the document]
chrisjurich: Sure. It has been changed.

enzy_htp/_interface/xtb_interface.py
```diff
@@ -8,28 +8,241 @@
...
+    QMOptimizationEngine,
+    QMResultEgg
+)
 
 
+from enzy_htp.core.job_manager import ClusterJob
 
 from plum import dispatch
+from enzy_htp import config as eh_config
+
```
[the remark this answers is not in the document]
chrisjurich: thanks man! the interface you defined was pretty easy to follow :)

enzy_htp/quantum/api.py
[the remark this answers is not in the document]
chrisjurich: I take that back. I think the `ElectronicStructure` should be returned. It is conceivable that you would want to do an optimization and then use the result of the calculation to evaluate some factor like dipole moment or whatever. I think doing the optimization in place makes sense or having an option to return a copy. Maybe return something like `List[Tuple[Structure, ElectronicStructure]]`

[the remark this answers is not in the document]
chrisjurich: Alternatively you could have the `Structure` object have an ElectronicStructure object that it owns

enzy_htp/_interface/xtb_interface.py
```diff
@@ -8,28 +8,241 @@
...
+                self.parent_interface.config()['XTB_EXE'],
+                run_info['args'])
+
+        os.chdir(start_dir)
+
+        self.parent_interface.update_coords(sr, run_info['expected_outfile'])
+
+        lines:List[str] = fs.lines_from_file(run_info['xtb_outfile'])
+
+        return  ElectronicStructure(
```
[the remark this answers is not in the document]
chrisjurich: See above comment but I think maybe a `Tuple` of `Strcucture` and `ElectronicStructure` would make more sense.

enzy_htp/structure/structure_constraint/__init__.py
```diff
@@ -10,14 +16,19 @@
     AngleConstraint,
     DihedralConstraint,
     ResiduePairConstraint,
+    )
+
+from .create_constraint import (
```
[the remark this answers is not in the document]
chrisjurich: Thanks!

enzy_htp/structure/structure_constraint/api.py
```diff
@@ -155,6 +166,13 @@ def check_consistent_topology(self) -> None:
         top = self.atoms[0].root()
         for atom in self.atoms:
             current_top = atom.root()
+           
+            if isinstance(current_top, ResidueCap):
+                current_top = current_top.parent                
+
+            if isinstance(current_top, StructureRegion):
+                current_top = curren_top.topology()
```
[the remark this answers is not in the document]
chrisjurich: lol XD

enzy_htp/structure/structure_constraint/api.py
```diff
@@ -155,6 +166,13 @@ def check_consistent_topology(self) -> None:
         top = self.atoms[0].root()
         for atom in self.atoms:
             current_top = atom.root()
+           
+            if isinstance(current_top, ResidueCap):
```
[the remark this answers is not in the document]
chrisjurich: So you will virtually always have at least a `DistanceConstraint` if you are doing a geometry optimization. I created some code which will lock all hydrogen bond distances, which may include `ResidueCap`'s. Also, you may want to do a `CartesianFreeze` on backbone atoms, which would include backbone heavy atoms, at least.

enzy_htp/structure/structure_constraint/api.py
```diff
@@ -155,6 +166,13 @@ def check_consistent_topology(self) -> None:
...
+                current_top = current_top.parent                
+
+            if isinstance(current_top, StructureRegion):
```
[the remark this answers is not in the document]
chrisjurich: In the current scheme, `StructureRegion` is the parent of the `ResidueCap`

enzy_htp/structure/structure_constraint/api.py
```diff
@@ -251,19 +275,17 @@ def __setitem__(self, key:str, value:Any ) -> None:
     # TODO move this to interface e.g.: RosettaInterface for current
     def score_energy(self) -> float:
```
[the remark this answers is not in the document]
chrisjurich: Sure.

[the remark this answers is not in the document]
chrisjurich: Moved to the `RosettaInterface`

enzy_htp/_config/modeller_config.py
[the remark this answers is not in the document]
chrisjurich: I'm working on it

enzy_htp/structure/structure_constraint/create_constraint.py
```diff
@@ -0,0 +1,310 @@
...
+        # atoms = stru_sele.select_stru(atoms).atoms
+        # TODO figure out how to resolve this loop importing issue
+        # structure_constraint -> structure_selection -> interface -> structure_constraint
+        # probably make 2 interface component 1 high 1 low?
+        raise Exception("TODO")
+    result = CartesianFreeze(atoms=atoms)
+    result.check_consistent_topology()
+    return result
+
+def create_backbone_freeze(stru: Structure, params:Dict) -> BackBoneFreeze:
```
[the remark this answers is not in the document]
chrisjurich: good catch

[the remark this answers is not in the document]
chrisjurich: Updated.

enzy_htp/structure/structure_constraint/create_constraint.py
chrisjurich: sure. done.

enzy_htp/structure/structure_operation/general.py
chrisjurich: Yeah this is a function I made that I forgot about. Basically I was finding a lot of errors because when `update_residues` is done it creates a deepcopy and will decouple `StructureConstraint`. I have since discovered a fix and do not use this.

enzy_htp/structure/structure_region/api.py
chrisjurich: I can tweak the code here a bit. In my experience this only applies to `ResidueCap` since they are not owned by the `Structure`. Let me know what to do.

enzy_htp/structure/structure_region/api.py
chrisjurich: I want to maintain the order of the `Residue()`'s. Since `not in` is slow-ish with a `list()`, I use the faster `set()` to check if the `Residue` has been added.

enzy_htp/structure/structure_region/capping.py
chrisjurich: I would prefer it stay this way. I make the chance because the atom ordering was getting messed up and would make it so that constraints would not work. It seems silly (and probably is) but when done this way, constraints are applied to the correct atoms.

enzy_htp/structure/structure_region/capping.py
chrisjurich: Whats the name of the typo? What's the address?

enzy_htp/structure/structure_region/capping.py
chrisjurich: Yes and yes

enzy_htp/_interface/amber_interface.py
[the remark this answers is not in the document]
shaoqx: You good! Thanks for fixing it.

test/structure/structure_region/test_capping.py
chrisjurich: This has been fixed! still works perfectly :)

enzy_htp/_interface/amber_interface.py
shaoqx: The goal is to make a new feature in PDBParser() (e.g.: PDBParser().get_structure(format="rosetta") and PDBParser().get_file_str(format="amber")) to canonicalize the atom naming. (e.g.: there will be an atom name mapper for each PDB format variant) We can add it after line 120 that built the Atom()s or after built the Structure() and in _write_pdb_atom(). 

We can have it in this PR or mark a TODO next to this function and make it in the next PR. ( I can help work on it )

enzy_htp/_interface/xtb_interface.py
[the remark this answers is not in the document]
shaoqx: Sounds good! I will change it when I make the `SimulationEngine`

enzy_htp/structure/structure_region/capping.py
shaoqx: I think it can be done with just cartesian coordinates. It is just transforming and rotating from different initial positions.

enzy_htp/quantum/api.py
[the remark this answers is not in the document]
shaoqx: Hmmmm yeah I can agree with that. Conventionally people use higher level of theory to do single point after opt to get the electronic structure but I agree in what we do, there will probably be the need of just getting energy/wavefunction after opt. I support your decision.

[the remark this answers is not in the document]
shaoqx: In this case, just returning the `ElectronicStructure` with an updated `.geometry` will be good. That covers all the need.

enzy_htp/structure/structure_constraint/api.py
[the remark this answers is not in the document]
shaoqx: What you might want is current_top.link_residue.root() I guess

enzy_htp/_interface/amber_interface.py
chrisjurich: Yes I like this design. I think ultimately we need a function that maps atom names that is then used in the parser, but can be used elsewhere also. If we had this function it could be useful to other workflows (i.e. `dock_reactants()` since you have to use Rosetta naming in that scheme.

enzy_htp/structure/structure_constraint/create_constraint.py
```diff
@@ -0,0 +1,310 @@
+"""
+
+Author: Qianzhen (QZ) Shao, <shaoqz@icloud.com>
+Author: Chris Jurich <chris.jurich@vanderbilt.edu>
+2024-01-01
```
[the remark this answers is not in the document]
shaoqx: I thought it should be 01-30 yesterday but it should be the time you make it yes

enzy_htp/structure/structure_region/capping.py
chrisjurich: I think I  misunderstood what you are saying. Yes you are right that the cap must "point" along 1,0,0. I thought this would be a good simplifying assumption to make the math easier, especially since there are only 1 or 2 capping groups even worth implementing after this

enzy_htp/structure/structure_constraint/api.py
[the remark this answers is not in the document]
shaoqx: It is not right now right? I saw it is set to ghost parent.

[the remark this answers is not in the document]
shaoqx: And I think it should not too logically. Because StructureRegion is not a real independent structure, it is under the context of a Structure() (it's topology). In this case, I think it is best to have ResidueCap have no parent and it can find the context by link_residue

enzy_htp/structure/structure_operation/general.py
shaoqx: Cool! We can remove it in this case

enzy_htp/structure/structure_region/api.py
shaoqx: Oh yes it only for the later comers after ctor. But still, ResidueCap atoms should also not be parented by the region but by the cap itself. see another related comment above.

enzy_htp/structure/structure_region/api.py
shaoqx: Sounds good

enzy_htp/structure/structure_region/api.py
shaoqx: Sounds good!

enzy_htp/quantum/api.py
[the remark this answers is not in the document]
chrisjurich: Great, I will do that!

enzy_htp/core/clusters/accre.py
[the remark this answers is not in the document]
chrisjurich: Pushed commit. Marking as resolved.

enzy_htp/chemical/residue.py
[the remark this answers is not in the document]
chrisjurich: Commit pushed. This should be resolved now.

enzy_htp/structure/structure_constraint/api.py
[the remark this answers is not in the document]
chrisjurich: Makes sense! I will implement this change

enzy_htp/structure/structure_constraint/api.py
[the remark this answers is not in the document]
chrisjurich: fixed and pushed.

enzy_htp/quantum/api.py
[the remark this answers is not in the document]
chrisjurich: Done!

enzy_htp/structure/structure_constraint/api.py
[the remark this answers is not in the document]
chrisjurich: Removed it!

enzy_htp/structure/structure_operation/general.py
chrisjurich: It has been removed! Sorry for the confusion

enzy_htp/structure/structure_region/api.py
chrisjurich: Fixed using your suggested trick! Thank you. I decided it was best to make a `ResidueCap.root()` method that redirects to the `ResidueCap.link_atom.root()`

enzy_htp/structure/structure_region/api.py
chrisjurich: So there were actually added after the `invovled_residues_with_free_terminal()` method was made. Basically there were some issues with how that code interacted with constraints (I think?) so I wanted to do it so that for each residue r1, r2, ... rn, the order of atoms would be r1_n_cap, r1, r1_c_cap, r2_n_cap, r2, r2_c_cap, etc. It's a little janky but the bugs went away.

enzy_htp/structure/structure_region/api.py
chrisjurich: It is gone, m'lord.

enzy_htp/structure/structure_region/capping.py
chrisjurich: It is now imported.

enzy_htp/structure/structure.py
chrisjurich: Nothing. I honestly don't remember why I put it. I'll remove lol.

enzy_htp/structure/structure_constraint/api.py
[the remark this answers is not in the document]
chrisjurich: I went ahead and changed it to that. Elegant solution you have proposed!

enzy_htp/structure/structure_region/api.py
chrisjurich: You are correct, and I think it is best for the `Atom()`'s in the `ResidueCap()` to be owned by the Cap and not the `StructureRegion`. I removed the above code that sets the `atom.parent.parent` to not `None`.

enzy_htp/structure/structure_region/capping.py
chrisjurich: I removed THe typo.

enzy_htp/structure/structure_region/capping.py
chrisjurich: Can you elaborate on this? I am confused what you mean.

enzy_htp/structure/structure_region/capping.py
chrisjurich: I agree this is vague. I have made it more specific/detailed now.

enzy_htp/structure/structure_region/capping.py
chrisjurich: Fixed! Pushed in commit

test/structure/structure_region/test_capping.py
chrisjurich: I have tried a few ways but i don't think there is a way to do this test without changing it a lot in the current scheme. thoughts?

enzy_htp/structure/structure_region/api.py
chrisjurich: I will switch it back

enzy_htp/structure/structure_region/capping.py
chrisjurich: I will change this back.

enzy_htp/structure/structure_region/api.py
chrisjurich: I deleted them.

enzy_htp/structure/structure_region/capping.py
chrisjurich: Changed back.

enzy_htp/structure/structure_region/capping.py
shaoqx: [after meeting]
This is needed by a temp fix from xtb interface for using .mol correctly.

test/structure/structure_region/test_capping.py
shaoqx: [after meeting]
Just need to create a ResidueCap() differently and test the deepcopy()

chrisjurich: Added this back in! Seems to be working ok now.

enzy_htp/_interface/amber_interface.py
shaoqx: Cool! I will make TODOs both here and pdb_io

enzy_htp/core/clusters/accre.py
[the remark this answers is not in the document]
shaoqx: I agree that this should be set elsewhere. I will add a TODO for it. This will be part of the config for the HPC of individual user.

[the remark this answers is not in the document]
shaoqx: Yeah I guess for ACCRE this is fine that I will open it for all ACCRE user. And it should not be changed.

enzy_htp/structure/structure_constraint/api.py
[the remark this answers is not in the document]
shaoqx: Good job!

enzy_htp/structure/structure_region/residue_caps.py
```diff
@@ -0,0 +1,579 @@
...
+            _LOGGER.error("link atom must be one of the link residue atoms")
+            raise ValueError
+
+        self._rtype = chem.ResidueType.RESIDUECAP
+        self._terminal_type = terminal_type 
+
+        if self.term_type == 'nterm':
+            self.atoms = self.get_nterm_atoms()
+        elif self.term_type == 'cterm':
+            self.atoms = self.get_cterm_atoms()
```
shaoqx, line 77: Have their parent ever been set to self?
You can use `self.set_children(atoms)` (this also set parent to self) and let `self.atoms` return `self.children` like Residue does.

enzy_htp/structure/structure_region/residue_caps.py
```diff
@@ -0,0 +1,579 @@
...
+        link/socket Atom()'s and link Residue() in a different Structure() with all of those
+        structural features. Typically used with a StructureEnsemble.
+
+        Args:
+            geom: The Structure() to apply the ResidueCap() to.
+
+        Returns:
+            A copy of the ResidueCap() specialization applied to the supplied Structure().
+        """
+        result = deepcopy(self)
```
shaoqx, line 95: I guess this can be type(self).__new__ or just use the entire constructor by type(self)(sth...)

shaoqx: deepcopy in this case is meaningless

enzy_htp/structure/structure_region/residue_caps.py
```diff
@@ -0,0 +1,579 @@
...
+        """Straightforward anchoring that is identical for both n-terminal and c-terminal capping."""
+        p1 = np.array(self.link_atom.coord)
+        p2 = np.array(self.socket_atom.coord)
+
+        direction = p2 - p1
+        direction /= np.linalg.norm(direction)
+    
+        cap_bond_distance = self.CAP_BOND_DISTANCE[self.link_atom.name]
+
+        self.atoms[0].coord = p1 + cap_bond_distance*direction
```
shaoqx, line 244: Be great if we can put these math in math_helper.

enzy_htp/structure/structure_region/residue_caps.py
```diff
@@ -0,0 +1,579 @@
...
+        direction /= np.linalg.norm(direction)
+        rot_mat = rotation_matrix_from_vectors(np.array([1,0,0]), direction)
+        cap_bond_distance = self.CAP_BOND_DISTANCE[self.link_atom.name]
+
+        for aa in self.atoms:
+            pt = np.array(aa.coord)
+            pt = np.transpose(np.matmul(rot_mat, np.transpose( pt  )))
+            pt += (direction*cap_bond_distance) + p1 
+            aa.coord = pt
+
```
shaoqx, line 326: Be great if we can put these math to math_hepler.py

enzy_htp/structure/structure_region/residue_caps.py
```diff
@@ -0,0 +1,579 @@
...
+        rot_mat = rotation_matrix_from_vectors(np.array([1,0,0]), direction)
+        
+        cap_bond_distance = self.CAP_BOND_DISTANCE[self.link_atom.name]
+
+        for aa in self.atoms:
+            pt = np.array(aa.coord)
+            pt = np.transpose(np.matmul(rot_mat, np.transpose( pt  )))
+            pt += (direction0*cap_bond_distance) + p0
+            aa.coord = pt
+
```
shaoqx, line 414: We would like to reuse these from math_helper as they are repetative for most anchor

test/structure/structure_region/test_capping.py
```diff
@@ -13,13 +14,195 @@
...
+        test_stru, "resi 2")
+    test_region = stru_regi.StructureRegion(atoms=sele.atoms)
+    unique_names = set()
+    for atom in test_region.atoms:
+        unique_names.add(atom.name)
+
+    assert len(test_region.atoms) == len(unique_names)
+        
+
+def test_all_cap_types_supported():
```
shaoqx, line 40: Please have the function name in these titles. I was not sure what this test is for until I read the code.

test/structure/structure_region/test_capping.py
```diff
@@ -13,13 +14,195 @@
...
+            test_region = stru_regi.StructureRegion(atoms=sele.atoms)
+            try:                
+                capping.capping_with_residue_terminals(test_region, nterm_cap=nterm_cap, cterm_cap=cterm_cap)
+            except Exception as exc:
+                assert False, f"capping_with_residue_terminals() failed with {nterm_cap=} {cterm_cap=}"
+
+
+
+
+def test_all_cap_types_unique_names():
```
shaoqx, line 59: same as above

test/structure/structure_region/test_capping.py
```diff
@@ -13,13 +14,195 @@
...
+         
+            for atom in test_region.atoms:
+                if not isinstance(atom.parent, ResidueCap):
+                    continue
+                link_distance:float=atom.parent.link_atom.distance_to(atom.parent.socket_atom)
+                assert link_distance >= 1.0 and link_distance <= 1.50
+
+
+
+def test_correct_num_caps():
```
shaoqx, line 100: same as above

shaoqx (changes requested):
Thank you for addressing all the comments! Great changes! There are several minor stuffs to fix before the merge!

shaoqx:
I found there is a bug when anchoring the NHCH3 cap. It makes `test/quantum/test_api.py::test_single_point_gaussian_lv2` fail.
Here is the structure region after capping
image

enzy_htp/structure/structure_region/residue_caps.py
chrisjurich: no. I will go ahead and do that!

chrisjurich: Fixed! It did not alter any of the tests in `test/structure/structure_region`

enzy_htp/structure/structure_region/residue_caps.py
chrisjurich: I refined this as much as possible and put a simpler implementation in!

enzy_htp/structure/structure_region/residue_caps.py
chrisjurich: See above comment, but same idea.

enzy_htp/structure/structure_region/residue_caps.py
chrisjurich: Basically I put a method in `StructureRegion` called `align_rigid_atoms()` which is called in all `ResidueCap` children except for `HCap` (for obvious reasons). I could not (or at least did not want to) put anything in `core` since the point is to operate on `Atom()` objects and obviously there are not `Atom()'s in `core`. I did move a function called `direction_unit_vector()` to `math_helper.py` since I was using that function a lot. There are like three other comments where I just reference this bigger comment, but the same approach is used elsewhere.

enzy_htp/structure/structure_region/residue_caps.py
chrisjurich: See above comments.

test/structure/structure_region/test_capping.py
chrisjurich: I tried to make them more descriptive. Hopefully these are good?

test/structure/structure_region/test_capping.py
chrisjurich: See above.

test/structure/structure_region/test_capping.py
```diff
@@ -13,13 +14,195 @@
...
+            capping.capping_with_residue_terminals(test_region, nterm_cap=nterm_cap, cterm_cap=cterm_cap)
+   
+            unique_names = set()
+            for atom in test_region.atoms:
+                unique_names.add(atom.name)
+            assert len(test_region.atoms) == len(unique_names)
+
+
+
+def test_all_cap_types_have_realistic_bond_distances():
```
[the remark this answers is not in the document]
chrisjurich: See above.

test/structure/structure_region/test_capping.py
chrisjurich: See above.

chrisjurich:
Third time is the charm? I went ahead and fixed the `ResidueCap.anchor()` problems, made that code more generalized, and also made the testing documentation more descriptive. Let me know if there is anything else but we should be ok! Everything passed the tests, including `test/structure/structure_residue`, `test/structure/structure_constraint`, and `test/quantum/test_api.py::test_single_point_gaussian_lv2`, which were noted as somewhat having issues.

enzy_htp/structure/structure_region/residue_caps.py
shaoqx: Cool!
But I think for these
```
            pt = np.transpose(np.matmul(rot_mat, np.transpose( pt  )))
            pt += (d0*bd) + p0
```
they can still go to core and be like apply rotation_matrix() or I bet you can find a function from `from scipy.spatial.transform import Rotation`  to do that

test/structure/structure_region/test_capping.py
shaoqx: I hope we can have a function name that is the target for this test noted either in the test function name or at least in docstring

enzy_htp/structure/structure_region/residue_caps.py
chrisjurich: Good point! We will create a more permanent solution at some point in the future. See issue https://github.com/ChemBioHTP/EnzyHTP/issues/164

chrisjurich: For reference I have tried to use `scipy.spatial.transform.Rotation` in the past but it was generally not very easy to use and I struggled to find a way to implement it especially given that our needs are not very robust.

shaoqx (approved):
Third time is the charm! Approved! Great changes!

PR #163, opened 2024-01-30
Chrisjurich/reduce pymol file io

chrisjurich:
Quick PR. It turns out that you can actually just use the pymol API function `read_pdbstr()` to load a `Structure()` without having to do file I/O. Updated it within `enzy_htp._interface.pymol_interface.load_enzy_htp_stru()`. Should be a quick PR!


shaoqx (approved):
Great PR! This is a good finding. I fixed the unit test and now it passes all of them for test_pymol_interface.py

PR #166, opened 2024-02-12
Shaoqz/feature multefdes

shaoqx:
# Add support of MultiMutDes-EF in position_pattern

- now we can use built-in functions to specify positions to mutate


chrisjurich (approved):
Record for the fastest PR.

PR #168, opened 2024-02-16
Bugfix: fixed the inconsistency of using the parser

shaoqx:
Fix a bug using .run()


chrisjurich (approved):
Thanks for fixing this! I appreciate it.

PR #169, opened 2024-02-17
2nd PR for MultiMutDes-EF

shaoqx:
# ModelingEngine
- Added a based ABC for modeling engine in general
- Added thermostability engine
- Half-done for RosettaCartesianddGEngine

# Mutation
- changed the import structure for Mutation class. It is needed by _interface.

# Analysis
- Added stability calculations
- Added a method for the coarse dEF calculation

# Workflow App
- added main script for MultiMutDes


enzy_htp/_interface/handle_types/modeling_engine.py
chrisjurich, line 1: Great! Thank you for making this

test/_interface/test_rosetta_interface.py
```diff
@@ -0,0 +1,43 @@
...
+from enzy_htp import interface, config
+
+DATA_DIR = f"{os.path.dirname(os.path.abspath(__file__))}/data/"
+WORK_DIR = f"{os.path.dirname(os.path.abspath(__file__))}/work_dir/"
+sp = PDBParser()
+ri = interface.rosetta
+
+def test_relax_w_stru():
+    """as said in the name. make sure relax works Structure() input."""
+    config._rosetta.RELAX = "$ROSETTA3/source/bin/relax.mpi.linuxgccrelease"
```
chrisjurich, line 23: Try using `config['rosetta.RELAX'] = /path/to/relax`

chrisjurich: Same for `MPI_EXECUTABLE` also

chrisjurich (changes requested):
A few small changes. Looks good, thanks for getting this PR setup! Will help me out a lot.

test/_interface/test_rosetta_interface.py
shaoqx: This way it doesn't autocomplete. I would argue for using the way I did as convention.

enzy_htp/_interface/handle_types/modeling_engine.py
shaoqx: Thank you too for bringing this up!

test/_interface/test_rosetta_interface.py
chrisjurich: The reason I opted to create the `[]` operator is:
   + Runs everything through a function so there is some degree of parsing/checking for correctness. Makes it so that the error message can be more descriptive also
   + makes it easier to update settings through datafiles like `~/.eh_config` or in workflow scripts

I am not opposed to supporting direct access, but would prefer to do it where there is not underscore since that looks weird and is generally not a good process. The two options are: 
   + provide support for the `__getattr__` so that when you call `config.rosetta` it is actually `config['rosetta']` **this is preferred IMO**
   + change all attributes from `_package` to `package`, (i.e. config._rosetta becomes config.rosetta); **this is less prefferred**

It may make the most sense to just punt this to an issue for later. I guess it is alright to make the convention to not use the brackets, but they are very nice for scripting/setting global configurations

shaoqx: Cool! I will apply it now.

shaoqx: I made them properties so that autocomplete works. The __getattr__ still can't work with autocomplete, which can't process code but can only read type hinting.

chrisjurich: Cool! As long as we keep both I am ok with it.

chrisjurich (approved):

chrisjurich:
Good job! Again, I appreciate the rebasing so that I can include the features in my own branch :)

PR #179, opened 2024-04-24
The PR for adding the template for the tutorial and bug fixes

shaoqx:
- Add a template for the tutorial
- Add functions in ARMer for re-run
- fix bugs

### DIFF
--- a/enzy_htp/core/file_system.py
+++ b/enzy_htp/core/file_system.py
@@ -161,11 +161,21 @@ def get_valid_temp_name(fname: str, is_symlink: bool = False) -> str:
     if is_symlink:
         while Path(result_fname).is_symlink():
             idx += 1
-            result_fname = f"{fname[:-len(suffix)]}_{str(idx)}{suffix}"
+            if len(suffix) == 0: # the fname is a dir
+                while fname.endswith("/"):
+                    fname = fname[:-1]
+                result_fname = f"{fname}_{str(idx)}{suffix}"
+            else:
+                result_fname = f"{fname[:-len(suffix)]}_{str(idx)}{suffix}"
     else:
-        while os.path.isfile(result_fname):
+        while os.path.isfile(result_fname) or os.path.isdir(result_fname):
             idx += 1
-            result_fname = f"{fname[:-len(suffix)]}_{str(idx)}{suffix}"
+            if len(suffix) == 0: # the fname is a dir
+                while fname.endswith("/"):
+                    fname = fname[:-1]
+                result_fname = f"{fname}_{str(idx)}{suffix}"
+            else:
+                result_fname = f"{fname[:-len(suffix)]}_{str(idx)}{suffix}"
     return result_fname
 
 
--- a/enzy_htp/core/file_system.py
+++ b/enzy_htp/core/file_system.py
@@ -164,18 +164,18 @@ def get_valid_temp_name(fname: str, is_symlink: bool = False) -> str:
             if len(suffix) == 0: # the fname is a dir
                 while fname.endswith("/"):
                     fname = fname[:-1]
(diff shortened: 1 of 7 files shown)


enzy_htp/core/file_system.py
chrisjurich, line 1: Can this be changed to format the temp name generation as `<name>_0001.txt`, `<name>_0002.txt`... etc.? I have found it kinda frustrating to see how many of something exist and padding it with zeros makes it easier to `ls` in directories and see what has been done. I think this would only entail changing line 178 from 
```
result_fname = f"{fname[:-len(suffix)]}_{str(idx)}{suffix}"
```
to 
```
result_fname = f"{fname[:-len(suffix)]}_{idx:04d}{suffix}"
```
and obviously the same change for the other parts.

enzy_htp/geometry/sampling.py
```diff
@@ -121,15 +121,15 @@ def equi_md_sampling(stru: Structure,
         temperature=[(0, 0), (0.05*0.9, prod_temperature), (-1, prod_temperature)],
         constrain=[freeze_backbone] + prod_constrain)
 
-    equi_step = parent_interface.build_md_step(
```
chrisjurich, line 124: good catch on the naming.

enzy_htp/geometry/sampling.py
```diff
@@ -344,7 +344,7 @@ def _serial_md_steps(
     results = []
     for i in range(parallel_runs):
         # create job path
-        sub_work_dir = f"{work_dir}/rep_{i}"
+        sub_work_dir = fs.get_valid_temp_name(f"{work_dir}/rep_{i}")
```
chrisjurich, line 347: Could we also do zero padding for `rep_{i}`? the change is only to `rep_{i:04d}'

template/template_hpc_submission.sh
```diff
@@ -16,11 +16,9 @@ source /home/shaoq1/bin/amber_env/amber-accre.sh
 # Multiwfn
 module load GCC/6.4.0-2.28  OpenMPI/2.1.1
 export KMP_STACKSIZE=200000000
-export Multiwfnpath=/home/jiany37/Software_installed/Multiwfn_3.7_bin_Linux_noGUI
+export Multiwfnpath=/home/shaoq1/bin/Multiwfn_3.7_bin_Linux_noGUI
```
chrisjurich, line 19: rip paul

template/template_main.py
chrisjurich, line 1: Good job cleaning this up. I think it is much easier to read and follow now

test/core/test_file_system.py
chrisjurich, line 1: This will need to be changed if we do the zero padding

chrisjurich (changes requested):
The only change that I request is padding out file names with zeroes. I think it makes it much easier to read and work with files during analysis.

enzy_htp/geometry/sampling.py
shaoqx: Changed!

enzy_htp/core/file_system.py
shaoqx: Changed!

template/template_hpc_submission.sh
shaoqx: lolll

template/template_main.py
shaoqx: :)

test/core/test_file_system.py
shaoqx: Just changed and passed

shaoqx:
Thank you! I have addressed all the comments!

chrisjurich (approved):

chrisjurich:
Looks good! Thanks for incorporating the proposed changes.

PR #182, opened 2024-06-06
Bugfix: fix 2 bugs about solvent and cache

shaoqx:
Fix 2 bugs:
1. bug discovered by @SwordJack about MDCRD parser
2. a potential caused by cache var

### DIFF
--- a/enzy_htp/_interface/amber_interface.py
+++ b/enzy_htp/_interface/amber_interface.py
@@ -2240,6 +2240,29 @@ class AmberInterface(BaseInterface):
         contents = "\n".join(contents)
         self.run_cpptraj(contents)
 
+    def remove_traj_solvent(
+            self,
+            traj_path: str,
+            prmtop_path: str,
+            out_path: str,
+            ) -> None:
+        """remove solvent in an Amber traj file using cpptraj
+        Args:
+            nc_path: The path to the .nc/.mdcrd file as a str().
+            prmtop_path: The path to the prmtop file as str().
+            out_path: the output MDCRD file path.
+        """
+        contents: List[str] = [
+            f"parm {prmtop_path}",
+            f"trajin {traj_path}",
+            "strip :WAT,Cl-,Na+",
+            f"trajout {out_path}",
+            "run",
+            "quit",
+        ]
+        contents = "\n".join(contents)
+        self.run_cpptraj(contents)
+
     # region == TODO ==
     def add_charges(self, stru: Structure, prmtop: str) -> None:
         """Method that adds RESP charges from a .prmtop file to the supplied Structure object. If the supplied prmtop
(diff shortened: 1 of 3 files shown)


SwordJack:
Hi Shao,

I've added round brackets to calls to the `stru.structures` method in `enzy_htp.analysis.spi_metric`. The test will raise following `TypeError` unless the round brackets are added.

```bash
>       for ss in stru.structures:
E       TypeError: 'method' object is not iterable

../../enzy_htp/analysis/spi.py:80: TypeError
========================================================================== short test summary info ==========================================================================
FAILED test_spi.py::test_spi_consistent_with_old_enzyhtp - TypeError: 'method' object is not iterable
======================================================================= 1 failed in 133.53s (0:02:13) =======================================================================
```

After the modification, the test is now able to be passed.

Best,
Zhong.

enzy_htp/_interface/amber_interface.py
```diff
@@ -2321,6 +2344,15 @@ def get_coordinates(self, mdcrd: str) -> Generator[List[List[float]], None, None
...
+
+        # remove solvent
+        if remove_solvent:
+            temp_dir = eh_config["system.SCRATCH_DIR"]
+            fs.safe_mkdir(temp_dir)
+            new_mdcrd = fs.get_valid_temp_name(f"{temp_dir}/mdcrd_parser_temp_rm_sol.mdcrd")
+            self.parent_interface.remove_traj_solvent(mdcrd, self.prmtop_file, new_mdcrd)
+            mdcrd = new_mdcrd
+
         with open(mdcrd) as f:
```
chrisjurich, line 2356: You can do this without having to use a while statement and it would be simpler I believe. you basically just want to parse in the whole thing as a giant buffer and read through it

enzy_htp/analysis/spi.py
```diff
@@ -47,7 +47,7 @@ def spi_metric(
         stru = stru.from_single_stru( stru )
 
     result = list()
-    for ss in stru.structures:
+    for ss in stru.structures():
```
chrisjurich, line 50: I'm not crazy about this change. Basically, the error originates from the cases when `get_itself` is the coord_parser given to the `StructureEnsemble` at construction. As such, I think it makes more sense to alter the `structures()` property/function in the class definition. .

test/analysis/test_spi.py
chrisjurich, line 1: Same comment as in `enzy_htp/analysis/spi.py`

chrisjurich (approved):
Looks good to me! Merge it in

enzy_htp/analysis/spi.py
chrisjurich: I take back what I said. At some point `structures()` was changed to not be a `property` and this attribute was not updated. Great catch!

PR #187, opened 2024-09-05
Shaoqz/hotfix dispatch kwarg

shaoqx:
Highlight and semi-fix a problem caused by Plum ignoring keyword arguments in dispatch
https://beartype.github.io/plum/keyword_arguments.html


shaoqx:
Also added an Common Errors & Solutions in the doc

PR #189, opened 2024-10-07
Shaoqz/bigblue supp

shaoqx:
Add support for BigBlue
Also fixed a bug in the Gaussian Interface


enzy_htp/structure/chain.py
chrisjurich, line 1: What lousy person put that there huh

setup.py
chrisjurich, line 1: good catch!

chrisjurich (approved):
Looks good! Go big blue.

enzy_htp/structure/chain.py
shaoqx: lol

PR #197, opened 2024-11-12
Cap and generate .ac file for modAA 

seb124:
This PR modifies the amber interface to generate an .ac file for a modified amino acid using antechamber. It first caps the N terminal with H and the C terminal with OH. Tests were also created for these functions.


shaoqx (changes requested):
Good job! Glad that we are going to have the support for ModAA! The capping function seems not working at the moment. Please address the above comments

test/structure/structure_region/test_capping.py
```diff
@@ -234,3 +234,19 @@ def test_capping_with_residue_terminals(helpers):
...
+
+    cterm_cap = "OH"
+    nterm_cap = "H"
+
+    try:                
+        capping.capping_with_residue_terminals(test_region, nterm_cap=nterm_cap, cterm_cap=cterm_cap)
+    except Exception as exc:
+        assert False, f"capping_with_residue_terminals() failed with {nterm_cap=} {cterm_cap=}"
+
+
```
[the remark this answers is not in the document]
shaoqx: Also add a test for the NHCH3 cap to see if it works.

enzy_htp/_interface/amber_interface.py
```diff
@@ -2519,12 +2524,18 @@ def antechamber_ncaa_to_moldesc(self,
...
+                out_path = f"{eh_config['system.NCAA_LIB_PATH']}/{ncaa.name}_{charge_method}-{gaff_type}.mol2"
 
         # 1. make ligand PDB
         temp_dir = eh_config["system.SCRATCH_DIR"]
         fs.safe_mkdir(temp_dir)
         temp_pdb_path = fs.get_valid_temp_name(f"{temp_dir}/{ncaa.name}.pdb")
+        if ncaa.is_modified():
+            # capping - cap C with OH and N with H
+            ncaa = create_region_from_selection_pattern(stru=ncaa, pattern="all", nterm_cap="H", cterm_cap="OH")
         pdb_io.PDBParser().save_structure(temp_pdb_path, ncaa)
```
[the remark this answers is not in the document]
seb124: Would the `convert_to_structure()` function be better placed in the `strucure.py` file? I see this `convert_res_to_structure` function at the end of the `strucure.py` file; I could add it underneath here
image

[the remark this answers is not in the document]
seb124: image
Looks like this already exists as the `topology` property in `structure_region.py`

shaoqx (changes requested):
Good job! Many problems resolved!
Please address the remains ones in the comment.

test/structure/structure_region/test_residue_caps.py
```diff
@@ -38,3 +39,87 @@ def test_residue_cap_deepcopy():
...
+    test_stru = sp.get_structure(f"{DATA_DIR}3FCR_modified.pdb")
+    protonate_stru(test_stru)
+    sele = stru_sele.select_stru(
+        test_stru, "resi 288")
+    test_region = stru_regi.StructureRegion(atoms=sele.atoms)
+    capping.capping_with_residue_terminals(test_region, nterm_cap='COCH3', cterm_cap='NHCH3')
+    # test
+    atoms = test_region.atoms
+    test_file = f"{WORK_DIR}test_capping_2.xyz"
+    answer_file = f"{DATA_DIR}answer_capping_2.xyz"
```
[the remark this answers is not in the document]
shaoqx: @seb124 can probably fix it using similar ways we did for OH cap

enzy_htp/_interface/amber_interface.py
```diff
@@ -2517,14 +2523,26 @@ def antechamber_ncaa_to_moldesc(self,
                           " ALWAYS check and explicit assign it using"
                           " Structure.assign_ncaa_chargespin()")
             raise ValueError
+
+        if ncaa.is_modified():
+            atom_type = "AMBER"
+            gaff_type = atom_type
```
shaoqx, line 2529: The gaff_type should not be changed. Our new parameter will still be a combination of the protein FF and GAFF for the modified side chain

enzy_htp/structure/structure_io/pdb_io.py
```diff
@@ -172,6 +172,18 @@ def get_file_str(cls, stru: Residue, if_renumber: bool = True, if_fix_atomname:
...
+    @dispatch
+    def get_file_str(cls, stru: ModifiedResidue, if_renumber: bool = True, if_fix_atomname: bool = True) -> str:  # pylint: disable=function-redefined
+        """
+        dispatch for supporting get pdb file str with Residue only
+        """
+        if if_renumber:
+            stru.parent.parent.renumber_atoms()
+        result_str = cls._write_pdb_residue(stru)
+        result_str += f"TER{os.linesep}END{os.linesep}"
+        return result_str
```
shaoqx, line 187: This seems not necessary since ModifiedResidue is a child class of Residue. Can you try if the test work w/o this function?

enzy_htp/structure/structure_io/xyz_io.py
```diff
@@ -0,0 +1,42 @@
+"""Generation/construction of Structure objects from PDB files and exporting it vice versa
+Definition of PDB file format (http://www.wwpdb.org/documentation/file-format-content/format33/v3.3.html)
+Also contains util function that requires knoweledge of PDB format
+
+Author: Qianzhen (QZ) Shao, <shaoqz@icloud.com>
```
shaoqx, line 5: lol change the file head and acknowledge yourself.

enzy_htp/structure/structure_io/xyz_io.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @classmethod
+    def make_xyz_from_atoms(self, file_path: str, atoms: List[Atom]):
+        lines = [
+            str(len(atoms)),
+            "",]
+        for at in atoms:
+            lines.append(f"{at.element} {at.coord[0]} {at.coord[1]} {at.coord[2]}")
+        fs.write_lines(file_path, lines)
+        return file_path
```
shaoqx, line 42: Please convert this to get_file_str.

You can use stru.atoms and do similar thing.

enzy_htp/structure/structure_region/residue_caps.py
```diff
@@ -531,6 +531,115 @@ def get_cterm_atoms(self) -> List[Atom]:
...
+        
+        link = np.array(self.link_atom.coord)
+        socket = np.array(self.socket_atom.coord)
+        
+        d0 = mh.direction_unit_vector(link, socket)
+
+        self.align_rigid_atoms(self.atoms, d0, link)
+
+    def net_charge(self) -> int:
+        return -1
```
shaoqx, line 628: The net charge should be 0

enzy_htp/structure/structure_region/structure_region.py
```diff
@@ -311,6 +318,65 @@ def topology(self) -> Structure:
...
+                else:
+                    aa.residue = aa.parent.link_atom.residue
+                if aa.residue not in res_to_cap_atoms:
+                    res_to_cap_atoms[aa.residue] = 0
+                res_to_cap_atoms[aa.residue] += 1
+
+            if aa.residue not in residue_mapper:
+                residue_mapper[aa.residue] = []
+            residue_mapper[aa.residue].append(aa)
+
```
shaoqx, line 344: Please use self.involved_residues and warp the mapper creation to a function called self.involved_residue_atom_mapper

shaoqx: Or another name if you have a better idea explaining its function.

enzy_htp/structure/structure_region/structure_region.py
```diff
@@ -311,6 +318,65 @@ def topology(self) -> Structure:
...
+        residues = []
+        for res, atoms in residue_mapper.items():
+            if res not in res_to_cap_atoms.keys():
+                res_to_cap_atoms[res] = 0
+            
+            if res.rtype == chem.ResidueType.MODIFIED:
```
shaoqx, line 350: Please use deepcopy() instead. And change the .atoms to the atoms in the mapper. This way all original attributes are kept.

shaoqx: If you made this by mimiking the way we construct in pdb_io, it is a great effort! However, we would like to treat this with fewer lines to make it more readable and extensible (i.e.: we may add new residue types)

enzy_htp/structure/structure_region/structure_region.py
```diff
@@ -311,6 +318,65 @@ def topology(self) -> Structure:
...
+                new_res = ModifiedResidue(int(res.idx), res.name, atoms, res.parent, net_charge=int(res.net_charge), multiplicity=int(res.multiplicity))
+                new_res.renumber_atoms(range(1, len(atoms) + res_to_cap_atoms[res] + 1))
```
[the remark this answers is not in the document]
shaoqx: What is the reason for using res_to_cap_atoms? I think len(atoms) already included the cap atoms right?

enzy_htp/structure/structure_region/structure_region.py
```diff
@@ -311,6 +318,65 @@ def topology(self) -> Structure:
...
+                residues.append(new_res)            
+            else:
+                new_res = Residue(int(res.idx), res.name, atoms, res.parent)
+                new_res.renumber_atoms(range(1, len(atoms) + res_to_cap_atoms[res] + 1))
+                residues.append(new_res)
+
+        chain_mapper: {Chain, List[Residue]} = {}
+        for res in residues:
+            if res.parent not in chain_mapper:
+                chain_mapper[res.parent] = []
```
shaoqx, line 370: You can use defaultdict(list) to simpify this part of the code. (i.e.: use it as in line 367 for chain_mapper)

enzy_htp/structure/residue.py
```diff
@@ -270,6 +275,14 @@ def n_side_residue(self)-> Union["Residue", None]:
...
         return result
+    
+    def find_idxes_atom_list(self, atom_idx_list: int) -> List[Atom]:
+        """find atom base on its idx. return a list reference of the atoms."""
+        result = []
+        atom_mapper = self.atom_idx_mapper
+        for idx in atom_idx_list:
+            result.append(atom_mapper[idx])
+        return result
     #endregion
```
shaoqx, line 286: Why do we need this?

shaoqx (changes requested):
Good job! Significant improvement have been made!
Please address the new comments!
Also a reminder about coding in an existing project: "always looks for existing functions before writing new lines"

enzy_htp/structure/residue.py
seb124: Not sure...must've added it while debugging/adding the convert_to_structure function

seb124: Oh okay, it is needed in the creation of a structure selection

enzy_htp/structure/structure_io/pdb_io.py
seb124: That's what I thought too, but without it, the test throws a plum.resolver.NotFoundLookupError

enzy_htp/structure/structure_io/xyz_io.py
seb124: Whoops haha

enzy_htp/structure/structure_region/structure_region.py
```diff
@@ -311,6 +318,65 @@ def topology(self) -> Structure:
...
+            chain_mapper[res.parent].append(res)
+        
+        chains = []
+        for chain, residues in chain_mapper.items():
+            chains.append(Chain(chain.name, residues))
+        
+        stru = Structure(chains)
+        stru.renumber_atoms()
+        return stru
```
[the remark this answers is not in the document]
seb124: Yeah, I updated it after I mentioned that to you, it should work for general StructureRegions at this point

enzy_htp/structure/structure_region/structure_region.py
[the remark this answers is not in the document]
seb124: You're completely right, after adding the new involved_residue_atom_mapper function I was able to remove this other Dict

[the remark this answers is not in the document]
seb124: However, it still seems I have to renumber them here due to an error in sorting when I attempt to simply renumber the structure at the end. This is because of the resiude cap atoms not being indexed yet

enzy_htp/structure/structure_io/pdb_io.py
shaoqx: What does the full content of the error look like? It might suggests a different problem.

enzy_htp/structure/structure_io/xyz_io.py
```diff
@@ -28,15 +30,19 @@ def get_structure(self, path: str) -> Union[Structure, tuple]:
         pass
 
     @classmethod
-    def get_file_str(self, stru: Structure) -> str:
-        pass
-
+    @dispatch
+    def get_file_str(self, stru: Structure) -> List[str]:
```
shaoqx, line 34: Please fix the type hinting of the return value

shaoqx: You can use Union[Structure, StructureRegion] to make one dispatched function works for both input types. Please give it a try.

test/structure/structure_region/test_capping.py
```diff
@@ -249,11 +249,8 @@ def test_capping_modaa(helpers):
...
     answer_file = f"{DATA_DIR}answer_capping_3.xyz"
-    atoms = test_region.atoms
-    file_path = f"{WORK_DIR}test_capping_3.xyz"
     
-    test_file = xyzp.make_xyz_from_atoms(file_path, atoms)
-    assert helpers.equiv_files(test_file, answer_file, consider_order=False)
+    result = xyzp.get_file_str(test_region)
 
-    fs.safe_rm(test_file)
+    assert fs.content_from_file(answer_file) == result
```
shaoqx, line 255: Can this test pass if you run it multiple times. Say 3 times.

enzy_htp/structure/structure_region/structure_region.py
```diff
@@ -325,50 +345,20 @@ def convert_to_structure(self, cap_as_residue = False) -> Structure:
...
-                residues.append(new_res)
+        residue_mapper = self.involved_residue_atom_mapper(cap_as_residue=cap_as_residue)
+
+        residues: List[Residue] = []
+        for res in self.involved_residues:
+            new_res = deepcopy(res)
+            new_res.atoms = residue_mapper[res]
+            new_res.chain = res.chain
+            new_res.renumber_atoms(range(1, len(new_res.atoms) + 1))
+            residues.append(new_res)
```
shaoqx, line 357: Instead of iterating the self.involved_residues, we would like to iterate the newly created residue_mapper.

shaoqx: Seems caps are left out in this case. I will explain it better in the comment for `involved_residue_atom_mapper`

enzy_htp/structure/residue.py
shaoqx: But in stru_selection.py, it is calling the one in structure.py right? Can you try removing this and see if there is any thing happen? Because based on a global search in the repo. This function is not called anywhere

enzy_htp/structure/structure_region/structure_region.py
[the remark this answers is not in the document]
shaoqx: Ahhh make sense

enzy_htp/structure/structure_region/structure_region.py
shaoqx: In renumber_atoms, you can use `num_atoms` method instead of len()

enzy_htp/structure/structure_region/structure_region.py
```diff
@@ -216,6 +216,26 @@ def atoms_by_residue(self) -> Dict[Residue, List[Atom]]:
         for atom in self.atoms:
             result[atom.parent].append(atom)
         return result
+    
+    def involved_residue_atom_mapper(self, cap_as_residue=False) -> Dict[Residue, List[Atom]]:
```
shaoqx, line 220: Good job! Only one thing about the intention of `cap_as_residue`:

- set to True: Every ResidueCap() itself will become a key of the dictionary. (because ResidueCap is also a decedent class of Residue). You might need to give them some missing properties of get_file_str needs them.
- set to False: Atom()s in caps are merged to the linked residues.

shaoqx:
BTW I helped fix the rot_vec_from_dihedral and the test. The only problem there is the axis direction vector is inverted. The direction is defined here https://en.wikipedia.org/wiki/Axis%E2%80%93angle_representation using the right hand convention.

enzy_htp/structure/residue.py
seb124: I get `AttributeError: 'ModifiedResidue' object has no attribute 'find_idxes_atom_list'`. I can change the type hinting on the function above this one, but it shouldn't be necessary considering ModifiedResidue is a child of Residue like you said

enzy_htp/structure/structure_region/structure_region.py
seb124: Sounds good, what should the parents of the ResidueCaps become?

test/structure/structure_region/test_capping.py
shaoqx: Cool! I was asking for the former case.

enzy_htp/structure/residue.py
shaoqx: Ahhh I see. I think the reason is that somewhere in your changes a ModifiedResidue (instead of Structure) is used as the input for the select_stru() function. 
In principle, we need to fix the wrong function call of select_stru() instead of here. Can you find out what test triggered such error?

enzy_htp/structure/structure_region/structure_region.py
shaoqx: It should be the same Chain() as it's linked residue.

enzy_htp/structure/residue.py
shaoqx: Ohhhh I see. It is this line 2545 in amber_interface.py, where we used ncaa as a Structure().
`ncaa_region = create_region_from_selection_pattern(stru=ncaa, pattern="all", nterm_cap="H", cterm_cap="OH")`

shaoqx: Since what we need is the ncaa under the Structure() context (for finding the socket atom for capping) not a new Structure() that only contain this ncaa, I suggest we add a new API for creating structure region from residues. I will add it tomorrow

test/structure/structure_region/test_residue_caps.py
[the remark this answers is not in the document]
shaoqx: converted to an issue #209

shaoqx (approved):
Amazing work! There are so many back and forth in this PR lol. Thank you! This will be a great aid to EnzyHTP.

PR #199, opened 2024-12-02
Yinjie: Fix `pdb_io` error handler.

SwordJack:
Dear colleagues,

This PR is to do some change to the behaviours after facing invalid PDB files. I want to `raise ValueError()` instead of `sys.exit(1)` when facing this issue, which will benefit to a smooth workflow in our Web Application.

Best,
Zhong.

### DIFF
--- a/enzy_htp/core/file_system.py
+++ b/enzy_htp/core/file_system.py
@@ -248,8 +248,9 @@ def check_file_exists(fname: str, exit_script: bool = True) -> None:
         _LOGGER.error(f"The file '{fname}' does not exist. Exiting...")
         exit(1)
     else:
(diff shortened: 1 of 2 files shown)


shaoqx (changes requested):
Good job! However, this change made some unit test not working. Please make sure all unit tests involving this change in test_pdb_io.py works

shaoqx:
You can do it by searching SystemExit in the file

SwordJack:
> You can do it by searching SystemExit in the file

You are right. This mismatch is fixed.

shaoqx (approved):
Good job! Merged

PR #201, opened 2024-12-06
Shaoqz/feature surface position

shaoqx:
New feature in pymol interface


SwordJack (approved):
Great job! We can use it to strengthen the function of mutant planner agent in EnzyHTP-GPT. Please update the agent's instruction and function tool_call config to enable the feature.

PR #203, opened 2024-12-17
Update: remove the hardcoded accre node exclude rule

shaoqx:
- add a keyword in the supported res_keywords


chrisjurich (approved):
LGTM. Thanks for the rapid fix haha

PR #214, opened 2025-02-28
Update main from develop

shaoqx:
We update the code in main since more and more people are using the code. This will more the tutorial more straight forward. (i.e.: no branch switching needed)

An update in the conda package will come right after.


shaoqx:
@chrisjurich This is merged right away for our collaborator.

PR #216, opened 2025-03-04
Generate .mc file for modAA

seb124:
This PR implements functions to create the mainchain file for modified amino acids. 

First, the connectivity of the structure containing the modAA is generated with `init_connectivity`. We then cap the modAA and pass the `StructureRegion` into the new `make_mc_file` function, found in `amber_interface`. This function finds the main chain in the modAA using the `ModifiedResidue` class's new function `find_mainchain` (which uses connectivity along with breadth-first-search to find the shortest path from the N-ter to the C-ter). `make_mc_file` also populates the .mc file with which atoms to omit (cap atoms) along with head/tail/pre_head/post_tail/charge data and returns the path to the new .mc file.

Unit tests were created for all of the above functions.


enzy_htp/structure/structure_region/structure_region.py
```diff
@@ -348,7 +348,7 @@ def convert_to_structure(self, cap_as_residue = True) -> Structure:
         residues: List[Residue] = []
 
         for res in residue_mapper:
-            new_res = deepcopy(res)
+            new_res = deepcopy(res) if not isinstance(res, ResidueCap) else res
```
shaoqx: This is not a desired action by of this function. They function works as converting a StructureRegion to a Structure in general. And the main different before and after this function for "self" is whether it knows the structural context. (i.e.: the returned Structure is an indivial Structure but not part of the original Structure)

shaoqx: Please change this back

enzy_htp/structure/modified_residue.py
```diff
@@ -81,6 +83,93 @@ def fix_atom_names(self) -> None:
                 _LOGGER.info(f"found atom with invalid name {atom}. changing it to {name}")
                 atom.name = name
 
+
+    def find_mainchain(self) -> List[Atom]:
```
shaoqx, line 89: Please keep this API consistent with the parent class i.e. Residue::mainchain_atoms

enzy_htp/structure/modified_residue.py
```diff
@@ -10,6 +10,8 @@
 
 from copy import deepcopy
 
+from enzy_htp.core.general import swapped_dict
```
shaoqx, line 13: Good job finding the function to use!

enzy_htp/_interface/amber_interface.py
```diff
@@ -2584,7 +2591,7 @@ def _generate_default_assigning_lines_for_build_md_parameterizer_or_step(paramet
         return result
 
     def antechamber_ncaa_to_moldesc(self,
-                                    ncaa: NonCanonicalBase,
+                                    ncaa: Union[NonCanonicalBase, StructureRegion],
```
shaoqx, line 3289: Good job! Make sure you also updated the docstring

enzy_htp/_interface/amber_interface.py
```diff
@@ -2624,11 +2638,15 @@ def antechamber_ncaa_to_moldesc(self,
...
         fs.safe_mkdir(temp_dir)
-        temp_pdb_path = fs.get_valid_temp_name(f"{temp_dir}/{ncaa.name}.pdb")
+
         if ncaa.is_modified():
-            # 1.1. Capping - cap C-terminal with OH and N-terminal with H
-            ncaa_region = create_region_from_residues(residues=[ncaa], nterm_cap="H", cterm_cap="OH")
-            ncaa = ncaa_region.convert_to_structure(cap_as_residue=False)
+            temp_pdb_path = fs.get_valid_temp_name(f"{temp_dir}/{ncaa.name}.pdb")
+        else:
+            temp_pdb_path = fs.get_valid_temp_name(f"{temp_dir}/{ncaa.name}.pdb")
```
shaoqx: What is this?

enzy_htp/_interface/amber_interface.py
```diff
@@ -2599,6 +2606,13 @@ def antechamber_ncaa_to_moldesc(self,
         Return:
             the out_path
             """
+
```
shaoqx: add a comment to declare the purpose of this section

let be `# init ncaa obj`

enzy_htp/_interface/amber_interface.py
```diff
@@ -2624,11 +2638,15 @@ def antechamber_ncaa_to_moldesc(self,
...
-            # 1.1. Capping - cap C-terminal with OH and N-terminal with H
-            ncaa_region = create_region_from_residues(residues=[ncaa], nterm_cap="H", cterm_cap="OH")
-            ncaa = ncaa_region.convert_to_structure(cap_as_residue=False)
+            temp_pdb_path = fs.get_valid_temp_name(f"{temp_dir}/{ncaa.name}.pdb")
+        else:
+            temp_pdb_path = fs.get_valid_temp_name(f"{temp_dir}/{ncaa.name}.pdb")
+
+        if ncaa.is_modified():
+            if maa_region:
+                ncaa = maa_region.convert_to_structure(cap_as_residue=False)
```
shaoqx, line 3343: add an else case where we keep the old logic. This way this function can also work with the raw maa as input.

shaoqx: by old logic I mean these:
```
            # 1.1. Capping - cap C-terminal with OH and N-terminal with H
            ncaa_region = create_region_from_residues(residues=[ncaa], nterm_cap="H", cterm_cap="OH")
            ncaa = ncaa_region.convert_to_structure(cap_as_residue=False)
```

enzy_htp/_interface/amber_interface.py
```diff
@@ -3286,6 +3304,51 @@ def make_mmpbgbsa_nc_file(
...
+        fs.write_lines(out_path, lines)
+        return out_path
+                
+
+        
+
+        
+        
+        
+
```
shaoqx: Please clean up blanks

enzy_htp/_interface/amber_interface.py
```diff
@@ -3286,6 +3304,51 @@ def make_mmpbgbsa_nc_file(
 
         self.convert_traj_to_nc(stru_esm.coordinate_list, temp_nc)
 
+    def make_mc_file(self, maa_region: StructureRegion, out_path: str):
+        """make the mc file for parameterization"""
+        maa: ModifiedResidue = maa_region.involved_residues[0]
```
shaoqx, line 4316: Give an error when the region contain more than one residues.

shaoqx: same for the moldesc function as well

enzy_htp/_interface/amber_interface.py
```diff
@@ -3286,6 +3304,51 @@ def make_mmpbgbsa_nc_file(
...
+        
+        if not maa.is_connected():
+            raise AttributeError(f"maa is not connected; use init_connectivity() first.")
```
shaoqx: Also use LOGGER to print out the warning

enzy_htp/_interface/amber_interface.py
```diff
@@ -3286,6 +3304,51 @@ def make_mmpbgbsa_nc_file(
...
 
+    def make_mc_file(self, maa_region: StructureRegion, out_path: str):
+        """make the mc file for parameterization"""
+        maa: ModifiedResidue = maa_region.involved_residues[0]
+        
+        if not maa.is_connected():
+            raise AttributeError(f"maa is not connected; use init_connectivity() first.") 
+
+        # main chain
+        mc_atoms = maa.find_mainchain()
```
shaoqx: As we will unify the API this function should return atoms including N and C. Change accordingly

enzy_htp/_interface/amber_interface.py
```diff
@@ -3286,6 +3304,51 @@ def make_mmpbgbsa_nc_file(
...
+            lines.append(f"MAIN_CHAIN: {aa.name}")
+
+        # omit cap atom names
+        for cap in maa_region.caps:
+            for aa in cap.atoms:
+                lines.append(f"OMIT_NAME: {aa.name}")
+
+        # find pre_head and post_tail atom types
+        for cap in maa_region.caps:
+            if cap.link_atom.element == "N":
```
shaoqx: We should determine the name of the link atom instead of element here. Imagine there are more than just C-ter and N-ter cap but also caps on other nitrogens.

enzy_htp/_interface/amber_interface.py
```diff
@@ -3286,6 +3304,51 @@ def make_mmpbgbsa_nc_file(
...
+                lines.append(f"PRE_HEAD_TYPE: {cap.socket_atom.element}")
+            
+        for cap in maa_region.caps:
+            if cap.link_atom.element == "C":
+                if cap.socket_atom.element != "N":
+                    _LOGGER.warning("Bond is not a classical peptide bond. MC generation may not work correctly.")
+                lines.append(f"POST_TAIL_TYPE: {cap.socket_atom.element}")
+
+        # charge of maa
+        lines.append(f"CHARGE: {maa.net_charge}")
```
shaoqx: Add this logic to the getter of these property in NonCanonicalBase

```
        if (ncaa.multiplicity is None) or (ncaa.net_charge is None):
            _LOGGER.error(f"supplied NCAA ({ncaa.name}) does not have charge and spin."
                          " ALWAYS check and explicit assign it using"
                          " Structure.assign_ncaa_chargespin()")
            raise ValueError

```

shaoqx: change the wording accordingly as it is no longer a supplied NCAA

shaoqx (changes requested):
Good job! We need some tweaks on the structure_region though.

enzy_htp/structure/modified_residue.py
```diff
@@ -81,6 +83,93 @@ def fix_atom_names(self) -> None:
...
+        return main_chain
+    
+    def clone_connectivity(self, other):
+        """clone connectivity from {other}.
+        The atoms are aligned and a connect record is added to each atom referencing
+        the connect from the aligned atom from other"""
+        # 1. san check
+        # - make sure other's atoms are not less than self's atoms
+
+        if len(self.atom_name_list) > len(other.atom_name_list):
```
shaoqx: We should not only determine the number of atoms but identity of atoms, that is, self.atom_name_list should be a subset of other's

enzy_htp/structure/modified_residue.py
shaoqx: lol I found it was from ncaabase but still keep looking for existing functions is always a good practice

enzy_htp/structure/modified_residue.py
shaoqx: Also please put the math component in core.math_helper

shaoqx: Good job implementing the algroithm! In practice we would like to use those existing libraries as much as possible. You might want to use the `networkx` library. (but you can put your code in the museum if you like to keep it lol see: https://github.com/ChemBioHTP/EnzyHTP/blob/e121cbf8c7dae24896e58a060d8d23fb192a9225/enzy_htp/core/general.py#L157)

enzy_htp/structure/structure_region/structure_region.py
shaoqx: I dont think change this back will affect anything as well the function is only used in the moldesc function

shaoqx: Update:
1. create deepcopy for ResidueCap that deal with socket_atom, link_atom/residue
2. create unit test and test the following three case:
       - copy a residue cap
       - convert_to_structure with cap_as_residue = True
3. modify convert_to_structure so that link information is restored

enzy_htp/structure/modified_residue.py
seb124: `is_subset` method

enzy_htp/structure/structure_region/structure_region.py
shaoqx: Also create two more test for convert_to_structure for cases involving caps and different cap treatments

like this
```
def test_convert_to_structure_cap():
    test_stru = sp.get_structure(f"{DATA_DIR}KE_07_R7_2_S.pdb")
    test_stru.assign_ncaa_chargespin({"H5J": (0,1)})
    test_stru_region = stru_regi.create_region_from_selection_pattern(
        test_stru, "resi 101+102"
    )
    test_stru_2 = test_stru_region.convert_to_structure()
    print(test_stru_2.residues)
```
You can think about what to assert

shaoqx: This test did not cause an infinite loop on my end. But definitely carried the wrong link_residue

enzy_htp/structure/modified_residue.py
seb124: Should this function set or simply return the mainchain atoms? I'll put the updated logic in the appropriate function in ModifiedResidue

enzy_htp/_interface/amber_interface.py
seb124: I found an issue with this - I get an incomplete residue error within capping_with_residue_terminals when I add this back, no matter what version of __deepcopy__ there is for ResidueCap. I will look more into the problem

enzy_htp/structure/structure_region/structure_region.py
```diff
@@ -348,11 +348,16 @@ def convert_to_structure(self, cap_as_residue = True) -> Structure:
         residues: List[Residue] = []
 
         for res in residue_mapper:
+            
+            if isinstance(res, ResidueCap):
+                link_residue = res.link_residue
+
             new_res = deepcopy(res)
             new_res.atoms = residue_mapper[res]
```
shaoqx: We are not copying the atoms, will this cause a reference problem? That is:
residue_mapper: res -> [atom0 -> res, atom1 -> res, ..., capatom0 -> cap, capatom1 -> cap, ...]
And for res it carry: res -> [atom0 -> res, atom -> res, ...]
But after the deepcopy it becomes:
new_res -> [atom0 -> new_res, atom1 -> new_res, ..., capatom0 -> new_res, capatom1 -> new_res, ...]
Would it cause:
res -> [atom0 -> new_res, atom -> new_res, ...]

Please make a unit test for convert_to_structure() and assert if the original residue have atoms that correctly points back to itself.

enzy_htp/structure/structure_region/structure_region.py
```diff
@@ -348,11 +348,16 @@ def convert_to_structure(self, cap_as_residue = True) -> Structure:
         residues: List[Residue] = []
 
         for res in residue_mapper:
+            
+            if isinstance(res, ResidueCap):
+                link_residue = res.link_residue
```
shaoqx: I cant see why we have these two lines

enzy_htp/structure/structure_region/structure_region.py
```diff
@@ -348,11 +348,16 @@ def convert_to_structure(self, cap_as_residue = True) -> Structure:
...
+
             new_res = deepcopy(res)
             new_res.atoms = residue_mapper[res]
 
             if isinstance(res, ResidueCap):
                 new_res.parent = res.link_residue.parent
+                new_res.link_residue = link_residue
```
shaoqx: The new link_residue should be the "new_res" version of the link_residue. That is, the deepcopied one

shaoqx: You can create a map that map the old link_residue to the new/copied one. This will be
1. iterate all caps to create the keys
2. during the main loop (this one), after the new_res generation process, if the old one is in the map, put the new one as the value

enzy_htp/structure/modified_residue.py
shaoqx: Good catch! I forgot I wrote these methods about mainchain atoms last summer. 

It seems we allowed user to self-assign the mainchain atoms with higher pirority. In this case, please keep the original function name, but call it in mainchain_atoms() in the `else` flow and assign _mainchain_atom with the return value & return it, and comment out the existing warning.

shaoqx: Also please update the docstring of mainchain_atom to state such user-assign/automation difference

shaoqx: Good job working with networkx!

enzy_htp/_interface/amber_interface.py
shaoqx: Could it be the way you test it? I didn't saw any unittest that would trigger this else flow

enzy_htp/_interface/amber_interface.py
shaoqx: Still, please use the mainchain_atom, but this time it is to allow user to customize

enzy_htp/structure/noncanonical_base.py
```diff
@@ -41,6 +41,13 @@ def __init__(self, residue_idx: int, residue_name: str, atoms: List[Atom], paren
     @property
     def net_charge(self) -> int:
         """Getter for the net_charge attribute."""
+
+        if self._net_charge is None:  
+            _LOGGER.error(f"NCAA does not have charge."  
+                          " ALWAYS check and explicit assign it using"  
+                          " Structure.assign_ncaa_chargespin()")  
+            raise ValueError
```
shaoqx: Let's return AttributeError instead. Same for multiplicity

enzy_htp/core/general.py
```diff
@@ -165,6 +165,50 @@ def num_ele_2d(list_2d: List[list]) -> int:
 #     return [[x] + y if x != GHOST_LIST_ELEMENT else y for x in curr_list
 #             for y in next_list]
 
+# The breadth-first-search algorithm I made for finding the mainchain of a modified residue
```
shaoqx: You can add you name or github id here haha

test/structure/structure_region/test_residue_caps.py
```diff
@@ -42,6 +42,24 @@ def test_residue_cap_deepcopy():
     assert ch3_copy.link_residue.parent is None
     assert ch3_copy.link_atom in ch3_copy.link_residue.atoms
```
shaoqx: This test is def wrong. Let's change it to a test that copy the entire structure region and see if the cap has new link_residue and is linked to it.

test/structure/structure_region/test_residue_caps.py
```diff
@@ -42,6 +42,24 @@ def test_residue_cap_deepcopy():
...
 
+def test_updated_residue_cap_deepcopy():
```
shaoqx: This is the right way to do the test.

test/structure/structure_region/test_structure_region.py
```diff
@@ -168,3 +168,10 @@ def test_convert_to_structure():
     test_stru_4 = test_stru_region_2.convert_to_structure()
     assert test_stru_3.is_same_topology(test_stru_4)
 
+    test_stru_5 = sp.get_structure(f"{DATA_DIR}KE_07_R7_2_S.pdb")
+    test_stru_5.assign_ncaa_chargespin({"H5J": (0,1)})
+    test_stru_region_3 = stru_regi.create_region_from_full_stru(
+        test_stru_5
+    )
```
shaoqx: This will not involve any cap. Please take a look at the PR comment for the example way of creating the region with cap

test/structure/structure_region/test_residue_caps.py
seb124: I implemented this new test in the last instance of the `test_convert_to_structure() `function

enzy_htp/structure/structure_region/structure_region.py
seb124: You're right, this did cause issues. I added a deepcopy for each atom addition in the `involved_residue_atom_mapper` function, and it seems to be fixed. Tested in `test_convert_to_structure`

enzy_htp/_interface/amber_interface.py
seb124: I think it had something to do with the convert_to_structure function, because after the fixes it works fine

enzy_htp/structure/structure_region/structure_region.py
```diff
@@ -226,11 +226,11 @@ def involved_residue_atom_mapper(self, cap_as_residue=True) -> Dict[Residue, Lis
         residue_mapper: {Union[Residue, ResidueCap], List[Atom]} = defaultdict(list)
         for res in self.involved_residues:
             for aa in res.atoms:
-                residue_mapper[aa.residue].append(aa)
+                residue_mapper[aa.residue].append(deepcopy(aa))
 
         for cap in self.caps:
             for aa in cap:
-                residue_mapper[cap].append(aa) if cap_as_residue else residue_mapper[aa.parent.link_atom.residue].append(aa) 
+                residue_mapper[cap].append(deepcopy(aa)) if cap_as_residue else residue_mapper[aa.parent.link_atom.residue].append(deepcopy(aa))
```
shaoqx: Good! But add an argument to trigger whether deepcopying atoms. (because this function can be used for other purpose)

enzy_htp/structure/structure_region/structure_region.py
```diff
@@ -344,20 +344,35 @@ def convert_to_structure(self, cap_as_residue = True) -> Structure:
...
+                new_res.atoms = residue_mapper[res]
+
                 new_res.parent = res.link_residue.parent
-            else:
-                new_res.parent = res.parent
-                
-            new_res.renumber_atoms(range(1, new_res.num_atoms + 1))
-            residues.append(new_res)
+                new_res.link_residue = old_to_new_res_mapper[res.link_residue]
+                if res.link_atom.name in [atom.name for atom in new_res.link_residue.atoms]:
```
shaoqx: Residue have a property called `atom_name_list`. use this one

test/structure/structure_region/test_residue_caps.py
```diff
@@ -37,10 +37,26 @@ def test_residue_cap_deepcopy():
         'cterm' 
     )
     ch3_copy = deepcopy(test_ch3)
-
     assert ch3_copy.link_residue is not test_res
-    assert ch3_copy.link_residue.parent is None
-    assert ch3_copy.link_atom in ch3_copy.link_residue.atoms
+
```
shaoqx: Good! We can just remove/replace the old test

enzy_htp/structure/structure_enchantment/connectivity.py
```diff
@@ -259,12 +260,16 @@ def _mol_desc_based_ncaa_method(ncaa: NonCanonicalBase, engine: str, ncaa_lib: s
...
+            mol_desc_path = f"{ncaa_lib}/{ncaa.involved_residues[0].name}_any.prepin"
+        else:
+            mol_desc_path = f"{ncaa_lib}/{ncaa.name}_any.prepin"  # swicth to mol2 after finish all unit tests of mol2_io
         MOL_DESC_GEN_MAPPER[engine](ncaa=ncaa, out_path=mol_desc_path)
     # 2. parse mol describing and clone into connectivity
     cnt_stru = MOL_DESC_PARSER_MAPPER[fs.get_file_ext(mol_desc_path)](mol_desc_path)
     cnt_ncaa = cnt_stru.residues[0]
-    ncaa.clone_connectivity(cnt_ncaa)
+
+    ncaa.clone_connectivity(cnt_ncaa) if not isinstance(ncaa, StructureRegion) else ncaa.involved_residues[0].clone_connectivity(cnt_ncaa)
```
shaoqx: Good! But put them in different lines to enhance readability

enzy_htp/_interface/amber_interface.py
```diff
@@ -324,6 +325,7 @@ def _parameterize_modified_res(self, maa: ModifiedResidue, gaff_type: str) -> Tu
         # init
         fs.safe_mkdir(self.ncaa_param_lib_path)
         target_method = f"{self.charge_method}-{gaff_type}"
+        maa_region = create_region_from_residues(residues=[maa], nterm_cap="H", cterm_cap="OH")
 
         # 0. search parm lib - same as ligand
         mol_desc_path, frcmod_path_list = search_ncaa_parm_file(maa,
```
shaoqx, line 356: This may not working well for the modified AA. Let's make a unit test for it after finishing the support for MAA. (it will be under test_ncaa_library.py)

shaoqx: To be clear, we finish all the following steps that give us frcmod file and do this

enzy_htp/_interface/amber_interface.py
```diff
@@ -336,11 +338,16 @@ def _parameterize_modified_res(self, maa: ModifiedResidue, gaff_type: str) -> Tu
...
         # 2.1 fix the wrong atom type given by antechamber
 
+
+        # create mc file
+        mc_path = f"{self.ncaa_param_lib_path}/{maa.name}_{target_method}.mc" # the search ensured no existing file named this
+        self.parent_interface.make_mc_file(maa_region=maa_region, out_path=mc_path)
+
         # 3. run prepgen on ac & mc get prepin
         prepin_path = fs.get_valid_temp_name(
             f"{self.ncaa_param_lib_path}/{maa.name}.prepin")
```
shaoqx: Let's also finish all the following steps in this PR so that we know if our .mc and .ac are legit

shaoqx (changes requested):
Good job! I can tell everything about mc is done! But let's finish the rest steps as well to make sure .mc is correct

shaoqx:
5.23 Meeting note:
To address the wrong atom type issue:
-  Use amber atom type for backbone atoms and CB. (to benefit from consistent psi/phi/etc. dihedral terms)
-  Use GAFF atom type for the rest.

.claude/CLAUDE.md
chrisjurich, line 1: Very cool to see that you are implementing support for claude!

enzy_htp/_interface/amber_interface.py
```diff
@@ -19,6 +19,7 @@
 from typing import Generator, List, Tuple, Union, Dict, Any
 from dataclasses import dataclass
 import pandas as pd
+from enzy_htp.structure.structure_region.structure_region import StructureRegion
```
chrisjurich: Do you have to go this deep or can it be imported from `enzy_htp.structure`? I am generally not sure but I think maybe this is possible. If not, dw.

enzy_htp/core/general.py
```diff
@@ -165,6 +165,50 @@ def num_ele_2d(list_2d: List[list]) -> int:
 #     return [[x] + y if x != GHOST_LIST_ELEMENT else y for x in curr_list
 #             for y in next_list]
 
+# Sebastian - The breadth-first-search algorithm I made for finding the mainchain of a modified residue
```
chrisjurich, line 171: What's going on here? If the code is not going to be in the final version, please delete.

enzy_htp/structure/modified_residue.py
```diff
@@ -49,12 +52,15 @@ def clone(self) -> ModifiedResidue:
...
+        """return a list of mainchain atoms. If the mainchain is already set, return those atoms. Otherwise, 
+           find the shortest path between N-term and C-term with find_mainchain(), then return it."""
         if len(self._mainchain_atoms) > 0:
             return self._mainchain_atoms
         else:
-            raise AttributeError(f"Missing mainchain information for {self.name} at {self.key_str}! "
-                                  "You can assign it using assign_mod_aa_mainchain()")
+            self._mainchain_atoms = self.find_mainchain()
+            return self._mainchain_atoms
+            # raise AttributeError(f"Missing mainchain information for {self.name} at {self.key_str}! "
```
chrisjurich: Delete this stuff that is commented out if you don't need it.

enzy_htp/structure/noncanonical_base.py
```diff
@@ -41,6 +41,13 @@ def __init__(self, residue_idx: int, residue_name: str, atoms: List[Atom], paren
     @property
     def net_charge(self) -> int:
         """Getter for the net_charge attribute."""
+
+        if self._net_charge is None:  
+            _LOGGER.error(f"NCAA does not have charge."  
+                          " ALWAYS check and explicit assign it using"  
+                          " Structure.assign_ncaa_chargespin()")  
+            raise AttributeError
```
chrisjurich: Please put the error code that you are logging here into the `AttributeError()`

enzy_htp/structure/noncanonical_base.py
```diff
@@ -51,6 +58,13 @@ def net_charge(self, val: int):
     @property
     def multiplicity(self) -> int:
         """Getter for the multiplicity attribute."""
+
```
chrisjurich, line 62: Same as above.

enzy_htp/structure/structure_region/residue_caps.py
```diff
@@ -226,6 +227,69 @@ def __str__(self) -> str:
         """str() rep of the ResidueCap() that shows its chemical type as well as what Atom()'s it is linked to."""
         return f"{self.cap_type}Cap(link_atom={self.link_atom.name},link_residue={self.link_residue.key_str})"
 
+    def __deepcopy__(self, memo: Union[Dict[int, Any], None] = None, _nil=[]):
```
chrisjurich, line 230: Is it necessary to list the `Args` here in the documentation? If this is some thing that you are just expected to do to support `deepcopy`, then don't worry. Just curious.

test/structure/structure_region/test_residue_caps.py
```diff
@@ -38,9 +38,10 @@ def test_residue_cap_deepcopy():
     )
     ch3_copy = deepcopy(test_ch3)
 
-    assert ch3_copy.link_residue is not test_res
-    assert ch3_copy.link_residue.parent is None
-    assert ch3_copy.link_atom in ch3_copy.link_residue.atoms
+    assert ch3_copy.link_residue is None
+    assert ch3_copy.link_atom is None
+    assert ch3_copy.socket_atom is None
+    assert len(ch3_copy.atoms) == len(test_ch3.atoms)
```
chrisjurich, line 46: also add `assert id(ch3_copy) != id(test_ch3)`

chrisjurich (changes requested):
@seb124 , this is a lot of good work! I have some minor issues that I bring up throughout. Note that since I have not been as involved in this PR, defer to @shaoqx , who seems to have guided this PR very well so far. Great work overall!

enzy_htp/core/general.py
shaoqx: This is the museum 🏛️

enzy_htp/_interface/amber_interface.py
```diff
@@ -3286,6 +3629,52 @@ def make_mmpbgbsa_nc_file(
...
+                lines.append(f"PRE_HEAD_TYPE {cap.socket_atom.element}")
+            
+        for cap in maa_region.caps:
+            if cap.link_atom.name == "C":
+                if cap.socket_atom.name != "N":
+                    _LOGGER.warning("Bond is not a classical peptide bond. MC generation may not work correctly.")
+                lines.append(f"POST_TAIL_TYPE {cap.socket_atom.element}")
+
+        # charge of maa
+        lines.append(f"CHARGE {maa.net_charge}")
```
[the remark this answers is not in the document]
shaoqx: no

shaoqx (approved):
Have this done. Tried Claude code. It helps a lot at the beginning stage but the sonnet-4 model seems have a hard time to clear all the bugs. Manually fixed everything and test.

shaoqx:
Only thing left is to expand the ff support in AmberInterface so that user is not too limited

enzy_htp/structure/structure_region/residue_caps.py
shaoqx: Yes, deepcopy is called in a recursive manner and memo in used by the built-in python deepcopy method

chrisjurich:
thanks for making the changes! LGTM! Approved

chrisjurich (approved):
Good job

PR #218, opened 2025-03-19
Shaoqz/feature new accre supp

shaoqx:
Add a new module for the new ACCRE on Rocky 9


SwordJack (approved):
Nice patch, solving the annoying troubles caused by ACCRE Rocky9 update.

PR #232, opened 2025-07-11
Yinjie/fix conda install

shaoqx:
Fix the junky environment.yml and fix the conda install


shaoqx:
This version of the yml took forever to solve environment. Trying two solutions:
- advice user to use mamba (this solution is unstable)
- record the solved environment and make the huge yml the old way (stable)

shaoqx:
Eventually I found new conda after 2023/11 use libmamba by default and is decently fast. Just added a check when user's conda is too old or otherwise configured.

